PR #37, opened 2015-07-14
Object equals

amerry:
#26 fixed one bug with operator== on json objects, but created another, because I didn't think through the behaviour of std::lower_bound properly.

This pull also includes an unrelated commit that makes the test suite build (did you fail to commit json_parser.hpp?)


danielaparker:
Thanks, Daniel

Issue #49, opened 2016-01-14
.get("key",stringval).as<std::string>(); fails


Bigpet:
on master this crashes for me on VS2013

    #include "jsoncons/json.hpp"
    #include <string>
    #include <iostream>
    
    using namespace jsoncons;
    using std::string;
    
    int main(int argc, char **argv)
    {
      json example;
      string result = example.get("test", string("too long string for short string")).as<string>();
      std::cout << "result:" << result << std::endl;
      return 0;
    }

Throws a `bad_alloc` on both master and the last tag during retrieving in `as<string>()`

Bigpet:
well `json::const_val_proxy` is move and copy constructible 

    std::cout << "copy_constructible " << std::is_copy_constructible<json::const_val_proxy>::value << std::endl;
    std::cout << "move_constructible " << std::is_move_constructible<json::const_val_proxy>::value << std::endl;

The issue is its reference `other.val_` could or could not point to `other.val2_`, so the only safe thing to do would be to copy `other.val_` to `this->val2_` like so:

     const_val_proxy(const const_val_proxy&rhs) : val2_(rhs.val_), val_(val2_){}; 

same for a move-constructor. Alternatively `const_val_proxy` could carry a flag whether or not it owns whatever val1_ points to.

I guess this is one of those cases where a non-owning reference is just as suble as a non-owning pointer

danielaparker:
Thanks for reporting this.

After revisiting get(name,default_val), I have to confess that I share your sentiment "the semantics of const_val_proxy are pretty confusing." I guess its original purpose was to delay the construction of the default json object, until it was determined that it was needed, but it was no longer doing that, if it ever did. So instead of taking your fix, I removed const_val_proxy altogether.

I had some difficulty getting a failing test case to test against, as VS 2015 was happy with your example. However, reverting to VS2013, the original code fails with your example, and the new code on master passes.

Thanks again,
Daniel

sjiekak:
I have a similar issue (vs2015) in the `jsoncons::create_impl` function during

`std::allocator_traits<Alloc>:: template rebind_traits<T>::construct(alloc, storage, std::forward<Args>(args)...);`

danielaparker:
Which version of jsoncons? Can you send me an example that shows the issue?

Daniel

sjiekak:
Version 0.99.1. 
I'll try to create a minimal example.

sjiekak:
[...]

PR #50, opened 2016-01-14
carify copy and move semantics of const_val_proxy

Bigpet:
related to #49

this is a pretty ugly "solution" but the semantics of `const_val_proxy` are pretty confusing to me.

Didn't run the test-suite on this one, so that the least that should be done.

There's probably a better solution than this but I'm not really familiar with the code-base and so the purpose of this struct sort of eludes me at the moment.

### DIFF
--- a/src/jsoncons/json1.hpp
+++ b/src/jsoncons/json1.hpp
@@ -1128,6 +1128,14 @@ public:
     private:
         const_val_proxy(); // no op
         const_val_proxy& operator = (const const_val_proxy& other); // noop
+        const_val_proxy(const const_val_proxy& other)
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks for submitting a fix.  

But after revisiting get(name,default_val), I have to confess that I share your sentiment "the semantics of const_val_proxy are pretty confusing." I guess its original purpose was to delay the construction of the default json object, until it was determined that it was needed, but it was no longer doing that, if it ever did. So instead of taking your fix, I removed const_val_proxy  altogether.

I had some difficulty getting a failing test case to test against, as VS 2015 was happy with your example. However, reverting to VS2013, the original code fails with your example, and the new code passes.

Thanks,
Daniel

PR #108, opened 2017-01-10
Filter objects extension

yplaksyuk:
Allows to apply filter expression over a single object. A single object might be considered as one-element array.

For example:
{
    "header": {
        "type": "A"
    }
}

The json-path expression "$.header[?(@.type == 'A')] allows to filter headers of type "A".


danielaparker:
Thanks for the contribution.

PR #114, opened 2017-03-09
Fix broken JSONCONS_NO_DEPRECATED

abbradar:
Also enable tests on Travis for this mode so this won't go unnoticed again, and fix style warnings of new GCC (about misleading identation).


danielaparker:
Thanks for both of these!

PR #116, opened 2017-06-01
Fix GCC 7 warnings

abbradar:
GCC 7 now warns if case fallthru is used without a marker (either C++17 attribute or a comment).


danielaparker:
These ones merged in, thanks.

PR #117, opened 2017-06-01
Convert variant_t to union type

abbradar:
Conversion is not as full as it could be with potential added safety.
We could e.g. define proper copy and move constructors in data_t.
However, this already is safer and shuts up GCC 7's warnings.

### DIFF
--- a/src/jsoncons/json.hpp
+++ b/src/jsoncons/json.hpp
@@ -129,7 +129,7 @@ public:
     typedef typename array::iterator array_iterator;
     typedef typename array::const_iterator const_array_iterator;
 
-    struct variant
+    union variant
     {
         struct base_data
         {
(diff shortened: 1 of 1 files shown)


abbradar:
Wait a bit on this one; I noticed new warnings when compiling with `-Og`.

abbradar:
Updated; now a full conversion into union. As a result a lot of code is cleaned up!

danielaparker:
The original implementation was a union. The motivation for changing the implementation to a variant type, essentially using placement new on json storage, was to support stateful allocators, in particular, the boost interprocess allocators. See the shared_memory example under more_examples. The problem is that for stateful allocators, `std::allocator_traits<T>::pointer` isn't necessarily a POD type, it isn't for the boost interprocess allocators, and becomes incompatible with unions. 

A second benefit of the variant implementation was to make better use of space for the small string optimization.

The main difference between the jsoncons variant and the boost or std::variant types is that the jsoncons one has a fixed number of alternatives, and uses a one byte discriminant, with other bytes used to store additional data.

Given that I think the variant implementation was correct in terms of alignment and so on, I would prefer to keep that implementation and fix up anything that's triggering GCC 7 warnings.

jeffreyscottgraham:
I agree with keeping the variant, though I would prefer std or boost variant.


On June 2, 2017 3:41:18 AM CDT, Daniel Parker <notifications@github.com> wrote:
>The original implementation was a union. The motivation for changing
>the implementation to a variant type, essentially using placement new
>on json storage, was to support stateful allocators, in particular, the
>boost interprocess allocators. See the shared_memory example under
>more_examples. The problem is that for stateful allocators,
>`std::allocator_traits<T>::pointer` isn't necessarily a POD type, it
>isn't for the boost interprocess allocators, and becomes incompatible
>with unions. 
>
>A second benefit of the variant implementation was to make better use
>of space for the small string optimization.
>
>The main difference between the jsoncons variant and the boost or
>std::variant types is that the jsoncons one has a fixed number of
>alternatives, and uses a one byte discriminant, with other bytes used
>to store additional data.
>
>Given that I think the variant implementation was correct in terms of
>alignment and so on, I would prefer to keep that implementation and fix
>up anything that's triggering GCC 7 warnings. 
>
>
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/danielaparker/jsoncons/pull/117#issuecomment-305726767

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

danielaparker:
Using `std::variant` would increase the size of the json value by 1/3, for storing extra data, e.g. precision for doubles. The current custom variant implementation has been here for six months, and closely follows similar implementations. I would like to see a list of the GCC 7 warnings.

abbradar:
@danielaparker I see, thank you for clarification! I don't have anything against your implementation as it is (I get the value of custom `variant` and if `union` can't be used I see no better way).

See [warnings](https://gist.github.com/abbradar/11cc6026bf6385b02f58da72f51c0655). TBH I didn't see anything incorrect in your implementation (if I remember C spec correctly type punning is OK for pointers to `char*`) and just thought that expressing that as more clean `union` and fixing the warning would kill two birds with one stone.

PR #120, opened 2017-07-26
Fix equality check of strings

jakalx:
In case one compares a string against something else for equality, the right
hand side was unconditionally converted to a string_view.

This change ensures that the right hand side is a string.


jakalx:
Resolves #121

danielaparker:
Thanks for this!

jakalx:
You are most welcome! I have to thank you!

PR #125, opened 2017-10-25
avoid clash with std::minus in VS2017

cenit:
while it works without any problem on gcc or clang, I have many problems with the "minus" label (json_parser.hpp) in Visual Studio, causing a name clash with std::minus (without ever using the std namespace)
If you know any better workaround, please let me know


danielaparker:
Strange, but thanks!

PR #129, opened 2017-11-28
Fix compile error from using custom allocator

xezon:
json_decoder fails to compile when using a custom allocator, because it tries to copy string types which use mismatching allocators.

Now that could be fixed differently if the json output is really meant to offer std::allocator only, but that is probably not desired, because a user naturally expects the out values with the specified allocator.


danielaparker:
Thanks for contributing!

PR #130, opened 2017-11-29
Ignore blank lines in csv files

xezon:
Blank lines, e.g. a blank line at file end, create an empty object in the json array, which is most likely not desired. Could be a csv option in the future.

### DIFF
--- a/include/jsoncons_ext/csv/csv_parser.hpp
+++ b/include/jsoncons_ext/csv/csv_parser.hpp
@@ -71,8 +71,8 @@ class basic_csv_parser : private parsing_context
     size_t index_;
     unsigned long column_;
     unsigned long line_;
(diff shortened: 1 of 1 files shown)


xezon:
Hmm apparently that fails test cases.

danielaparker:
I'll have a look

xezon:
I tested the setting `ignore_empty_values` and it still does creates an empty object in the array. I will rethink the logic based on your comment.

danielaparker:
Try current master. I've added a `csv_parameter` `ignore_empty_lines` which defaults to `true`. I've included two test cases in csv_tests.cpp with default `ignore_empty_lines` and `ignore_empty_lines` set to false.

xezon:
Works nicely. Thank you.

PR #131, opened 2017-12-01
Refactor json variant data

xezon:
I actually only want to use jsoncons, but I felt this change is necessary for 2 reasons:
1) removing all these reinterpret_cast's which make me very nervous
2) make the json value much easier to inspect in debugger

I haven't been able to compile the jsoncons tests locally. I am using cmake + vs 2017 and the compilation fails with error MSB1009, ALL_BUILD.vcxproj missing

Careful inspection is required. I hope I made no mistake.

### DIFF
--- a/include/jsoncons/json.hpp
+++ b/include/jsoncons/json.hpp
@@ -166,7 +166,7 @@ public:
         {
             json_type_tag type_id_;
 
-            base_data(json_type_tag id)
+            explicit base_data(json_type_tag id)
                 : type_id_(id)
             {}
         };
@@ -174,7 +174,7 @@ public:
         class null_data : public base_data
         {
         public:
-            null_data()
+            explicit null_data()
                 : base_data(json_type_tag::null_t)
             {
             }
@@ -183,7 +183,7 @@ public:
         class empty_object_data : public base_data
         {
         public:
-            empty_object_data()
+            explicit empty_object_data()
                 : base_data(json_type_tag::empty_object_t)
             {
             }
@@ -193,12 +193,12 @@ public:
         {
             bool val_;
         public:
-            bool_data(bool val)
+            explicit bool_data(bool val)
                 : base_data(json_type_tag::bool_t),val_(val)
             {
             }
 
-            bool_data(const bool_data& val)
+            explicit bool_data(const bool_data& val)
                 : base_data(json_type_tag::bool_t),val_(val.val_)
             {
             }
@@ -214,12 +214,12 @@ public:
         {
             int64_t val_;
         public:
-            integer_data(int64_t val)
+            explicit integer_data(int64_t val)
                 : base_data(json_type_tag::integer_t),val_(val)
             {
             }
 
-            integer_data(const integer_data& val)
+            explicit integer_data(const integer_data& val)
                 : base_data(json_type_tag::integer_t),val_(val.val_)
             {
             }
@@ -234,12 +234,12 @@ public:
         {
             uint64_t val_;
         public:
-            uinteger_data(uint64_t val)
+            explicit uinteger_data(uint64_t val)
                 : base_data(json_type_tag::uinteger_t),val_(val)
             {
             }
 
-            uinteger_data(const uinteger_data& val)
+            explicit uinteger_data(const uinteger_data& val)
                 : base_data(json_type_tag::uinteger_t),val_(val.val_)
             {
             }
@@ -255,14 +255,14 @@ public:
             uint8_t precision_;
             double val_;
         public:
-            double_data(double val, uint8_t precision)
+            explicit double_data(double val, uint8_t precision)
                 : base_data(json_type_tag::double_t), 
                   precision_(precision), 
                   val_(val)
             {
             }
 
-            double_data(const double_data& val)
+            explicit double_data(const double_data& val)
                 : base_data(json_type_tag::double_t),
                   precision_(val.precision_), 
                   val_(val.val_)
@@ -288,7 +288,7 @@ public:
         public:
             static const size_t max_length = (14 / sizeof(char_type)) - 1;
 
-            small_string_data(const char_type* p, uint8_t length)
+            explicit small_string_data(const char_type* p, uint8_t length)
                 : base_data(json_type_tag::small_string_t), length_(length)
             {
                 JSONCONS_ASSERT(length <= max_length);
@@ -296,7 +296,7 @@ public:
                 data_[length] = 0;
             }
 
-            small_string_data(const small_string_data& val)
+            explicit small_string_data(const small_string_data& val)
                 : base_data(json_type_tag::small_string_t), length_(val.length_)
             {
                 std::memcpy(data_,val.data_,val.length_*sizeof(char_type));
@@ -325,11 +325,12 @@ public:
             typedef typename std::allocator_traits<Allocator>:: template rebind_alloc<string_storage_type> string_holder_allocator_type;
             typedef typename std::allocator_traits<string_holder_allocator_type>::pointer pointer;
 
-            pointer ptr_;
+            pointer ptr_ = nullptr;
 
             template <typename... Args>
             void create(string_holder_allocator_type allocator, Args&& ... args)
             {
+                JSONCONS_ASSERT(ptr_ == nullptr);
                 typename std::allocator_traits<Allocator>:: template rebind_alloc<string_storage_type> alloc(allocator);
                 ptr_ = alloc.allocate(1);
                 try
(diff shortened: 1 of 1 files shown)


include/jsoncons/json.hpp
```diff
@@ -325,11 +325,12 @@ class basic_json
             typedef typename std::allocator_traits<Allocator>:: template rebind_alloc<string_storage_type> string_holder_allocator_type;
             typedef typename std::allocator_traits<string_holder_allocator_type>::pointer pointer;
 
-            pointer ptr_;
+            pointer ptr_ = nullptr;
 
             template <typename... Args>
             void create(string_holder_allocator_type allocator, Args&& ... args)
             {
+                JSONCONS_ASSERT(ptr_ == nullptr);
```
xezon: I added an assert and 0 initialized variant pointers. This should add a bit more confidence that the setup is correct.

include/jsoncons/json.hpp
```diff
@@ -654,21 +658,88 @@ class basic_json
...
+            integer_data integer_data_;
+            uinteger_data uinteger_data_;
+            double_data double_data_;
+            small_string_data small_string_data_;
+            string_data string_data_;
+            byte_string_data byte_string_data_;
+            array_data array_data_;
+            object_data object_data_;
+            empty_object_data empty_object_data_;
+        };
```
xezon: I removed the aligned storage and replaced it with union members of all the different variant types. If this project was targeted for c++17, we could use std::variant, but this should do for c++11.

A programmer can now inspect the variant data conveniently in debugger.

include/jsoncons/json.hpp
```diff
@@ -654,21 +658,88 @@ class basic_json
...
+        template <> uinteger_data&     as<uinteger_data>()     { return uinteger_data_; }
+        template <> double_data&       as<double_data>()       { return double_data_; }
+        template <> small_string_data& as<small_string_data>() { return small_string_data_; }
+        template <> string_data&       as<string_data>()       { return string_data_; }
+        template <> byte_string_data&  as<byte_string_data>()  { return byte_string_data_; }
+        template <> array_data&        as<array_data>()        { return array_data_; }
+        template <> object_data&       as<object_data>()       { return object_data_; }
+        template <> empty_object_data& as<empty_object_data>() { return empty_object_data_; }
+
+        template <class T> const T& as_c() const;
```
xezon: This is a bit odd. This template function should be simply named `as() const`, which is the counter part to the above non const template member function. But for some weird reason this would completely break IntelliSense in VisualStudio 2017, meaning no more syntax highlighting / auto completion / etc in the entire project.

So I named it `as_c() const` for now.

include/jsoncons/json.hpp
```diff
@@ -1129,679 +1083,84 @@ class basic_json
...
+    private:
+        template <class TypeA, class TypeB>
+        void swap_a_b(variant& other)
+        {
+            TypeA& curA = as<TypeA>();
+            TypeB& curB = other.as<TypeB>();
+            TypeB tmpB(std::move(curB));
+            other.construct_var<TypeA>(std::move(curA));
+            construct_var<TypeB>(std::move(tmpB));
+        }
```
xezon: The swap function was very huge. I simplified it into 3 template functions. I think that is easier to read and requires little maintenance.

xezon:
Thank god it built :)

include/jsoncons/json.hpp
```diff
@@ -759,210 +839,93 @@ class basic_json
...
         {
-            new(reinterpret_cast<void*>(&data_))array_data(val);
+            construct_var<array_data>(val);
         }
         variant(const array& val, const Allocator& alloc)
         {
-            new(reinterpret_cast<void*>(&data_))array_data(val,alloc);
-        }
-        template<class InputIterator>
-        variant(InputIterator first, InputIterator last, const Allocator& a)
```
xezon: I removed this template function because it is not used and uses a constructor in array_data that does not exist.

danielaparker:
The original implementation was a union, like most of the other implementations such as nlohmann that followed jsoncpp, rapid json and json spirit. 

The motivation for changing the implementation to a variant type, essentially using placement new on json storage, was to support stateful allocators, in particular, the boost interprocess allocators. See the shared_memory example under more_examples. The problem is that for stateful allocators, `std::allocator_traits<T>::pointer` isn't necessarily `T*` anymore, or POD, in particular, it isn't for the boost interprocess allocators, which are incompatible with unions. The shared_memory example won't work with the union implementation.

A second benefit of the custom variant implementation is to make better use of space for the small string optimization.

The main difference between the jsoncons variant and the boost variant or `std::variant` types is that the jsoncons one has a fixed number of alternatives, and uses a one byte discriminant, with other bytes used to store additional data. For us, it wouldn't be practical to use `std::variant`, as that would increase the size of each json value by 1/3, for space to store extra data, e.g. precision for doubles. Otherwise they're analogous, and the jsoncons implementation follows conventional techniques for implementing variant types, it's safe and portable. This implementation has been in effect for over a year, and no issues have been reported on any of the environments where jsoncons is used, Windows, Linux, MacBook and Android.

The swap function when `std::allocator_traits<T>::pointer` is POD (which is the one your compiler would be using in your code) is actually one line:

            std::swap(data_,other.data_);

The more complicated swap function only applies when `std::allocator_traits<T>::pointer` is not POD, such as the boost interprocess allocators. 

Full support for stateful allocators has it's challenges, and we still have some work to do, json parsing and decoding needs to be looked at to make sure everything uses the right allocator. But I don't want to take a step backward. While I don't expect many users will want to use the boost interprocess allocators, they're actually an excellent test, because if you're accidentally allocating _anything_ with `std::allocator` by mistake,  it's not going to make it's way to another process.

xezon:
Any chance to make the json value debug friendly somehow however? It's really hard to inspect it, because without casting to the right type it will only show gibberish. With this change I am able to inspect json values.

I am not aware about the implications you listed. Union is still constructed with placement new in this code. So I am not sure what you mean !?

xezon:
I gave this more thought and can't think of any reason this should not work. It should behave the exact same way as before, but cleaner. If I made a mistake somewhere, it should be fixable. Did you actually inspect/run the code or just assumed it does not work?

Note:
Placement new on std::aligned_storage and union is the exact same thing. Why do you think it behaves different?

danielaparker:
array_data and object_data cannot be part of a union if they are not POD. Try compiling and running the shared memory example for an example where they would not be POD. 

    On Friday, December 1, 2017 12:37 PM, xezon <notifications@github.com> wrote:
 

 I gave this more thought and can't think of any reason this should not work. It should behave the exact same way as before, but cleaner. If I made a mistake somewhere, it should be fixable. Did you actually inspect/run the code or just assumed it does not work?Note:
Placement new on std::aligned_storage and union is the exact same thing. Why do you think it behaves different?—

danielaparker:
Regarding debug view, it may be worth exploring adding a .natvis improving the debug view of json in the MSVC GUI.

xezon:
"array_data and object_data cannot be part of a union if they are not POD"

I don't understand? The union lays out storage just like `std::aligned_storage` does. Thus if it works for this std::aligned_storage object it must work for union storage too.

I ran the shared_memory.cpp test and it ran fine. Do you have a code that is supposed to break it?

xezon:
Can you please revisit this change? I am not convinced your assessment is correct.

danielaparker:
Which part of my assessment? I'm correct that the boost interprocess stateful allocator that use shared memory  would result in string_data, array_data and object_data being not POD (plain old data.)

You're correct that my statement about unions and non-POD are no longer accurate, since C++ 11. Up until C++ 11, it was illegal to include non-POD members in a union. Since C++ 11, this has been relaxed. See e.g.
[Questions regarding C++ non-POD unions](https://stackoverflow.com/questions/19764150/questions-regarding-c-non-pod-unions) and [What are Unrestricted Unions proposed in C++11?](https://stackoverflow.com/questions/5548636/what-are-unrestricted-unions-proposed-in-c11).

I'm still not convinced that it would be a good idea to use a union here.

xezon:
The argument should be the other way around, where the std::aligned_storage is being questioned. The union guarantees type and size safety, the aligned storage does not. Granted, I could not see a mistake in your implementation, but that does not prove the approach as superior method. This change also cuts like 700 lines of boiler plate code.

Natvis for aligned storage is probably possible, not sure. This change requires no natvis for inspection, although a natvis could greatly improve it by only adding the right type to view.

include/jsoncons/json.hpp
danielaparker: Right, I've removed this function from master

danielaparker:
It took a while for me to come around, but I've incorporated most of this pull request into v0.145.1. My main concern was that the function
```c++
json_type_tag type_id() const
{
        return base_data_.type_id_;
}
```
is UB according to the Standard. However, by eliminating the base class, and duplicating type_id in each data class, it appears we meet the requirement of common initial sequence for the data classes as defined in the Standard.

PR #135, opened 2017-12-01
Fix compile error from using JSONCONS_HAS_STRING_VIEW

xezon:
std::istringstream does not take a std::string_view (vs 2017)

### DIFF
--- a/include/jsoncons_ext/csv/csv_parser.hpp
+++ b/include/jsoncons_ext/csv/csv_parser.hpp
@@ -784,7 +784,7 @@ private:
             {
             case csv_column_type::integer_t:
                 {
-                    std::istringstream iss(value);
+                    std::istringstream iss(value.data());
                     int64_t val;
                     iss >> val;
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks for trying that!  I still haven't used a compiler that supports `std::string_view`, so haven't tried that. Passing `value.data()` isn't quite right though, as string views don't guarantee that `data()` is null terminated. Could you try

 std::istringstream iss(std::string(value));

instead? I missed that the `std:;string` constructor that takes a string view is explicit.

Thanks,
Daniel

xezon:
True. This does not look right. `std::string(data, size)` should be the right thing.

danielaparker:
They should be equivalent. std::string has an explicit constructor that takes a `std::string_view`, which would use data() and size(). Either is fine.

xezon:
Yes, but on the other hand std::string does not take the jsoncons custom string view type, so you would have to rely on implicit cast.

PR #140, opened 2017-12-23
Fix mismatching allocator types

xezon:
New master branch does not compile on MSVC c++17

The allocator types are mismatching. I fixed all errors I could find, but there might be more, as I might not have all templates in my compile unit. I suggest to pass allocators as templated templates. Then it is easier to use them for various types.

### DIFF
--- a/include/jsoncons/json_decoder.hpp
+++ b/include/jsoncons/json_decoder.hpp
@@ -18,7 +18,7 @@
 
 namespace jsoncons {
 
-template <class Json,class Allocator=std::allocator<typename Json::char_type>>
+template <class Json,template <class Type> class Allocator=std::allocator>
 class json_decoder : public basic_json_input_handler<typename Json::char_type>
 {
 public:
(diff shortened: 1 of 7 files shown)


danielaparker:
Thanks for raising this issue.

Following boost and other examples, however, the template parameter should be just class `Allocator = std::allocator<ValueType>` rather than `template<class U> class Allocator = std::allocator`. When we use the allocator with multiple value types, we are supposed to rebind the allocator as needed for every allocation, so the initial `ValueType` of the allocator isn't important. We generally use `CharT`, could even use `void`, except not all compilers support `std::allocator<void>`.

I've changed the code to rebind the Allocator template parameter to the `value_type` for the standard library containers wherever they appear in the code.

Having a template parameter for each value type would, I think, be impractical. There are too many - we have `char_allocator_type`, `string_allocator_type`, `csv_mode_allocator_type`, `csv_type_info_allocator_type`, and `string_vector_allocator_type` in `basic_csv_parameters` alone. Besides, using a vector of `csv_type_info` is an internal detail that could change, so we don't want to expose that to the user.

xezon:
Ok nice. I will just pull your fix then.

danielaparker:
I'll be grateful for feedback, as the compilers I'm using don't give me any diagnostics for the mismatch, so I could easily have missed one or two.

Thanks,
Daniel

PR #164, opened 2018-09-24
Use double type for 1st argument to ldexp()

jpritikin:
Spotted by Prof Brian Ripley <ripley@stats.ox.ac.uk>


danielaparker:
Thanks!

PR #165, opened 2018-10-01
Fix infinite recursion with bignums and misc. undefined behavior

Cebtenzzre:
Hi,
I discovered this project due to its inclusion in [miloyip/nativejson-benchmark](https://github.com/miloyip/nativejson-benchmark).  I locally updated its submodule for jsoncons from commit 88194db to the latest master.  After adjusting the tests to the current API, the benchmark segfaulted due to a stack overflow during the conformance testing.

Long story short, with the help of gdb, I found a case where conversion of `basic_json` values backed by bignums to primitive types entered an infinite recursion -- for example, `as_double()` parses the string representation of the value, expecting it to convert to a double, but its data type remains a string because it is semantically a bignum and too big to fit in a double.  I fixed this by using `as_bignum()` and a conversion operator if the semantic type matched `bignum_tag`.

I found undefined behavior with ubsan and msan along the way -- `neg_` was sometimes not initialized, and `data_` is allocated with `std::allocator::allocate()` and deallocated with `delete[]`, even though [`allocate()` uses `::operator new()`](https://en.cppreference.com/w/cpp/memory/allocator/allocate) (without the brackets) internally.  I changed it to `deallocate()` for the sake of consistency.

### DIFF
--- a/include/jsoncons/bignum.hpp
+++ b/include/jsoncons/bignum.hpp
@@ -233,6 +233,7 @@ public:
 
     basic_bignum(unsigned short u)
     {
+        neg_ = false;
         initialize_from_integer( u );
     }
 
@@ -245,6 +246,7 @@ public:
 
     basic_bignum(unsigned int u)
     {
+        neg_ = false;
         initialize_from_integer( u );
     }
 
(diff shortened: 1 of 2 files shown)


danielaparker:
Thanks for reporting these. Before bignum was introduced, integers that overflowed were handled as doubles, but I hadn't thought through the implication for `as_double()`, `as_int64()`, etc. when that was changed to bignum.

PR #167, opened 2018-11-14
__FILE__ and __LINE__ macroses removed from release version

zhskyy:
these macroses are generates a literal constants that contains full file names and after that the release exe file contains these literal constants with my local directories structure.

so it should be removed from release build and looks  the JSONCONS_ASSERT macros also may be fully removed from release but I don't sure.

let me know if You have any question

thanks for this cool library.


danielaparker:
No objection to the change itself, but a lot of failing tests.

https://travis-ci.org/danielaparker/jsoncons/jobs/455067714
https://ci.appveyor.com/project/danielaparker/jsoncons/builds/20300162

PR #170, opened 2018-12-19
added `sum()` and `mult()` functions.

soberich:
Hi, 

I want to further create functions (especially simple sum() would be helpful) which will work NOT ONLY inside filter. 
Could you point out where to start with and are there some gotchas.
I want to do simply `$sum($..someArrayOfDoubleValues)` to get a number or, to get an array (as you always get here) with size of 1 and `arr[0]` equal to what was calculated.

BTW. 
In one of the test I wrote this 
`std::string path = "$.store.book[?(@.price > sum($.store.book[*].price) / $.store.book.length)].title"; // unexpected exception if '$.store.book.length'.`
Is it expected?

### DIFF
--- a/doc/ref/jsonpath/jsonpath.md
+++ b/doc/ref/jsonpath/jsonpath.md
@@ -19,11 +19,11 @@ Stefan Goessner's javascript implemention returns `false` in case of no match, b
 
 Unlike XML, the root of a JSON text is usually an anonymous object or array, so JSONPath identifies the outermost level of the text with the symbol $.
 
-JSONPath expressions can use the dot–notation
+JSONPath expressions can use the dot-notation
 
     $.store.book.0.title
 
-or the bracket–notation 
+or the bracket-notation 
 
 
     $[store][book][0][title]
@@ -94,12 +94,15 @@ Precedence|Operator|Associativity
 
 #### Aggregate Functions
 
-Functions can be called inside filter expressions. The input to an aggregate function is the value of a JSONPath expression.
+Functions can be called __*only inside filter expressions*__. The input to an aggregate function is the value of a JSONPath expression.
+There are not much use cases for aggregation functions like `mult()` inside filter expressions but still they could be used, here are examples:
 
 Function|Description|Result|Example
 ----------|--------|-------|---
-min()|Provides the minimum value of an array of numbers|double|$.store.book[?(@.price < max($.store.book[*].price))].title
-max()|Provides the maximum value of an array of numbers|double|$.store.book[?(@.price > min($.store.book[*].price))].title
+`min(...)`|Provides the minimum value of an array of numbers|`double`|`$.store.book[?(@.price < max($.store.book[*].price))].title`
+`max(...)`|Provides the maximum value of an array of numbers|`double`|`$.store.book[?(@.price > min($.store.book[*].price))].title`
+`sum(...)`|Provides the sum value of an array of numbers|`double`|`$.store.book[?(@.price > sum($.store.book[*].price) / 4)].title`
+`mult(...)`|Provides the multiplication value of an array of numbers|`double`|`$.store.book[?(479373 < mult($..price) && mult($..price) < 479374)].title`
 
 ### Examples
 
--- a/include/jsoncons_ext/jsonpath/jsonpath_filter.hpp
+++ b/include/jsoncons_ext/jsonpath/jsonpath_filter.hpp
@@ -35,6 +35,8 @@ JSONCONS_DEFINE_LITERAL(ampamp_literal,"&&")
 JSONCONS_DEFINE_LITERAL(pipepipe_literal,"||")
 JSONCONS_DEFINE_LITERAL(max_literal,"max")
 JSONCONS_DEFINE_LITERAL(min_literal,"min")
+JSONCONS_DEFINE_LITERAL(sum_literal,"sum")
+JSONCONS_DEFINE_LITERAL(mult_literal,"mult")
 
 template<class Json>
 struct PathConstructor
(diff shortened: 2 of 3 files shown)


danielaparker:
Thanks for the contribution, appreciated. I'll have a look at the test case you singled out, length is a special case, so may not currently be handled.

Daniel

soberich:
Hey Daniel, 
I want this PR to be a starter to actually create aggregation functions which work not only inside filter expressions.
Would be great to have, let's say, a sum of all book prices like.
`$.sum($.store.book.price)` (or some other notation, say `$.store.book.price.sum()` ...)
and other functions like
`$.mult($.store.book.price)`
`$.avg($.store.book.price)`
I double write this now as I am not sure I expressed this intend in my original pr description clear enough.
I will appreciate if you may guide a little where to look for such handles in code (depending on your available time whether 2 sentences or a hundred sentences will be helpful).
Have an active project which may use the lib if such things would've be there.
Thanks.

danielaparker:
I understand that. But let's make that a separate PR :-)
I think I prefer sum($.store.book.price) over $.store.book.price.sum(), as more consistent with functions in the filters, and more general. The argument would be a path that would evaluate to a set of nodes (could be generalized to take more arguments, but later), and would return a set of nodes. The path could continue after the function invocation, e.g. sum($.store.book.price)[<filter expression>]


    On Thursday, December 20, 2018, 4:06:00 a.m. EST, soberich <notifications@github.com> wrote:  
 
 
Hey Daniel,
I want this PR to be a starter to actually create aggregation functions which work not only inside filter expressions.
Would be great to have, let's say, a sum of all book prices like.
$.sum($.store.book.price) (or some other notation, say $.store.book.price.sum() ...)
and other functions like
$.mult($.store.book.price)
$.avg($.store.book.price)
I double write this now as I am not sure I expressed this intend in my original pr description clear enough.
I will appreciate if you may guide a little where to look for such handles in code (depending on your available time whether 2 sentences or a hundred sentences will be helpful).
Have an active project which may use the lib if such things would've be there.
Thanks.

danielaparker:
I would suggest doing this in two steps. 

First, modify the state transitions in `json_query.hpp` such that any valid function name (starts with a character, contains no spaces, etc.) evaluates as an identity function. 

Second, implement specific functions like sum.

So to start, the default section in
```c++
            case path_state::start: 
                switch (*p_)
                {
                case ' ':case '\t':
                    break;
                case '$':
                case '@':
                    {
                        string_type s;
                        s.push_back('$');
                        node_set v;
                        v.emplace_back(std::move(s),std::addressof(root));
                        stack_.push_back(v);

                        state_ = path_state::expect_dot_or_left_bracket;
                    }
                    break;
                default:
                    err_handler_->fatal_error(jsonpath_parser_errc::expected_root, *this);
                    ec = jsonpath_parser_errc::expected_root;
                    return;
                };
                ++p_;
                ++column_;
                break;
```
should transition to a new state, `path_state::function_name`, if `*p_` is alphabetic.

Once you have the function name, and a '(', you need the argument (a path expression), and when you have ')', you can evaluate the node set with
```c++
    jsonpath_evaluator<Json,const Json&,detail::VoidPathConstructor<Json>> evaluator;
    evaluator.evaluate(root,buffer.data(),buffer.length());
   // JSON result = evaluator.get_values();
```
   Instead of calling get_values(), you could copy the contents of the evaluator stack_ (pointers)
   to the  current stack_. Then change the state
```
    state_ = path_state::expect_dot_or_left_bracket;
```
And that should give you an identity function. (Step 1)

i.e. 

 `myfunc($.store.book.price)`

should return the same as `$.store.book.price`

Step 2 will be to implement specific functions.

danielaparker:
Hi soberich,

I missed one  "gotcha" in my comments above. I said "you should copy the contents of the evaluator stack_ (pointers into the root, will remain valid even when evaluator is gone)
to the current stack_". While it's true that pointers to json values selected from the root will remain valid, the set of selected nodes can also include pointers to temporary json values (e.g. as a result of evaluating "length"),  which were held in an array belonging to the `jsonpath_evaluator`, and therefore had the lifetime of the evaluator. I fixed that by having one vector of unique_ptrs to the temporary json values, and passing it down through all the function calls. It's only purpose is to keep these pointers alive until `json_query` or `json_replace` returns. 

To make sure this all works as it's supposed to, I implemented the state transitions for the identity function, up to this point in `json_query`, line 602, on master:
```c++
case path_state::function_argument:
    switch (*p_)
    {
    case ')':
    {
        jsonpath_evaluator<Json, JsonReference, PathCons> evaluator;
        evaluator.evaluate(root, buffer_.data(), buffer_.length(), temp_json_values);

        //std::cout << function_name << "\n";
        //std::cout << buffer_ << "\n";
        // Here you can lookup the function name
        // If there is no such function, it's an error
        // err_handler_->fatal_error(jsonpath_errc::expected_root, *this);
        // ec = jsonpath_errc::invalid_function_name;
        // return

        // identity function
        for (node_set& ns : evaluator.stack_)
        {
            // ns.val_ptr is a pointer to a Json value in the result set
            // So you can call functions ns.val_ptr->is_double() etc.
            // compute a result, and put the result node(s) on the stack
            stack_.push_back(ns);
        }
        state_ = path_state::expect_dot_or_left_bracket;
        break;
    }
    default:
        buffer_.push_back(*p_);
        break;
    }
    ++p_;
    ++column_;
    break;
```
I've added a test case in jsonpath_function_tests.cpp. I'll leave "step 2" for you.

Daniel

danielaparker:
Just FYI, I took your proposal a little further, and functions are now supported consistently over JSONPath expressions.

danielaparker:
Regarding

```
$.store.book[?(@.price > sum($.store.book[*].price) / $.store.book.length)].title"
```
I don't think JSONPath allows for referring to the root element directly in a filter. I've added a `count` function so that you can write
```
$.store.book[?(@.price > sum($.store.book[*].price) / count($.store.book[*]))].title"
```
We could also add an `evaluate` function (following xpath) that would allow
```
$.store.book[?(@.price > sum($.store.book[*].price) / evaluate($.store.book.length))].title"
```
I've changed the name of your function `mult` to `prod`, for consistency with a similar function in matlab and numpy.

Again, your contribution and proposal very much appreciated, 
Daniel

PR #189, opened 2019-10-15
Fix compiler (gcc 9.2) warning "accessible non-virtual destructor"

KonstantinPlotnikov:
GCC 9.2 produces warning: `‘class jsoncons::json_exception’ has virtual functions and accessible non-virtual destructor`


danielaparker:
Thanks for this!

KonstantinPlotnikov:
Thank you for so fast and friendly json library! It helped me a lot.

PR #215, opened 2020-02-11
Fix incorrect default mapping with headers in docs.

dpsi:
The mapping `n_columns` does not exist.


danielaparker:
Thanks! There's one more problem with that line, it should be the other way around,  i.e., defaults to `n_objects`, otherwise `n_rows`. Could you fix that as well before I accept the request?

dpsi:
Here you go!

danielaparker:
Thanks, appreciated,
Daniel

PR #219, opened 2020-04-02
Added a fuzzer and integration with oss-fuzz.

DavidKorczynski:
I have integrated jsoncons into OSS-fuzz (https://github.com/google/oss-fuzz). This is a free service offered by Google where they perform continuous fuzzing of open-source projects that are used by the industry and with the only expectation that maintainers of the project will use and patch the bugs found by OSS-Fuzz. If you are interested in this, could you please provide me with an email address that will be receiving the bug reports that my fuzzers find? If you are happy with this I can also continue developing more fuzzers so the project gets a good coverage. OSS-Fuzz already has several json parsers! 

David

### DIFF
--- a/fuzzers/fuzz_parse.cpp
+++ b/fuzzers/fuzz_parse.cpp
@@ -1,0 +1,15 @@
+#include <jsoncons/json_parser.hpp>
+#include <jsoncons/json.hpp>
+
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks! Email address is danielaparker@yahoo.com.

PR #221, opened 2020-04-06
Added fuzzers to fuzz extensions.

DavidKorczynski:
I added three more fuzzers to hit the entry points of the extensions. With these we should be able to get a good amount of coverage on the code!

### DIFF
--- a/fuzzers/fuzz_bson.cpp
+++ b/fuzzers/fuzz_bson.cpp
@@ -1,0 +1,20 @@
+#include <jsoncons/json.hpp>
+#include <jsoncons_ext/bson/bson.hpp>
+#include <sstream>
+#include <vector>
+#include <utility>
+#include <ctime>
+#include <limits>
+#include <catch/catch.hpp>
+
+using namespace jsoncons;
+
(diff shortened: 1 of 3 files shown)


danielaparker:
Thanks! I can see the value of fuzzers.

Incidentally, I received a notification yesterday form fuzz with the link

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21589

but I am unable to access the link (permission denied). Any suggestions?

DavidKorczynski:
I think the issues is likely to be your account is not set up as a google account? You need to register it as such and then login using that account - the account that you login with should be the email given here: https://github.com/google/oss-fuzz/blob/master/projects/jsoncons/project.yaml

Perhaps this solves it?

danielaparker:
I think I need to use a google account to get full access. Can you change the email address to 

danielaparker@gmail.com

DavidKorczynski:
> I think I need to use a google account to get full access. Can you change the email address to
> 
> [danielaparker@gmail.com](mailto:danielaparker@gmail.com)

Fixed and merged into oss-fuzz!

PR #223, opened 2020-04-10
Updated fuzzers to catch exceptions thrown by the json parsing itself.

DavidKorczynski:
Added some bug fixing that catches exceptions thrown by the parsers themselves. This should fix various bug reports from OSS-Fuzz.

### DIFF
--- a/fuzzers/fuzz_cbor.cpp
+++ b/fuzzers/fuzz_cbor.cpp
@@ -15,6 +15,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
                 json j2 = decode_cbor<json>(is);
         }
         catch(jsoncons::ser_error e) {}
+        catch(jsoncons::json_runtime_error<std::runtime_error> e2) {}
 
         return 0;
 }
--- a/fuzzers/fuzz_csv.cpp
+++ b/fuzzers/fuzz_csv.cpp
@@ -14,8 +14,8 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
                 csv::csv_reader reader1(input, decoder, options);
                 reader1.read();
         }
-        catch(jsoncons::ser_error e) {}
-        catch(jsoncons::json_runtime_error<std::runtime_error> e) {}
-
+        catch (jsoncons::ser_error e) {}
+        catch (jsoncons::json_runtime_error<std::runtime_error> e) {}
+        catch (json_runtime_error<std::invalid_argument> e3) {}
         return 0;
 }


danielaparker:
Thanks. I can now access the reports, so will start going through them.

DavidKorczynski:
Cool - also, feel free to reach out if you have questions. Also, I intend to push some more work on the project to make it well covered with fuzzers so if you have any priorities on how I code them up then please let me know!

danielaparker:
It would be useful to have a separate pure parser test for each of json, csv, bson, cbor, msgpack, ubjson, e.g.
```
default_json_visitor visitor;
 json_reader reader(is, visitor);

std::error_code ec;
reader.read(ec);
```
(A `default_json_visitor` discards the parse events produced by the parser.)  This allows us to test the parser independently of the decoder.

danielaparker:
Then to test parser+encoder,
```c++
std::string s;
json_string_encoder encoder(s);
json_reader reader(is, encoder);

std::error_code ec;
reader.read(ec);
```
This just pumps the parse events directly into the encoder.

DavidKorczynski:
Ah! So the following works for me:
For cbor and binary format parsing:
```c++
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
{
        std::string s(reinterpret_cast<const char*>(data), size);
        std::istringstream is(s);
        std::vector<uint8_t> s1;
        cbor_bytes_encoder encoder(s1);
        cbor_reader reader(is, encoder);

        std::error_code ec;
        reader.read(ec);

        return 0;
}
```

and for json parsing:

```c++
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
{
        std::string s(reinterpret_cast<const char*>(data), size);
        std::istringstream is(s);
        json_string_encoder encoder(s);
        json_reader reader(is, encoder);

        std::error_code ec;
        reader.read(ec);

        return 0;
}
```

is this what you had in mind?

danielaparker:
Right. That will exercise the encoders directly from the parsed input.  

I fixed Issue 21589: jsoncons:fuzz_parse: Stack-overflow in `jsoncons::basic_json`. For deeply nested `basic_json` objects or arrays, the C++ compiler has an internal stack size which limits how far it can recurse down when deleting from the top. I fixed that by flattening the hierarchy using the heap and deleting sequentially.

In general, stack overflows are a big concern. The json/csv/bson/cbor/msgpack/ubjson parsers do not recurse on the stack, so should be okay. The encoders, on the other hand, do recurse on the stack, as currently written, I would expect some clusterfuzz failures . I'm going to add a `max_depth` option parameter for all the parsers/encoders, and give a default.

The jsonpointer and jsonpath `unflatten` functions also recurse on the stack. I'll add a `max_depth` parameter for those.

I notice we're getting some out of memory errors for parsing binary formats. I think what's happening is that some input looks to the parser as a field with a very large size, and when jsoncons reserves memory for the field that follows, it fails. I think it would be okay to catch `std::bad_alloc` in the fuzzer.  

Incidentally, I would suggest

```try {} catch(const jsoncons::ser_error& e){}```    
rather than

```try {} catch(jsoncons::ser_error e){}```    

Although not an issue with `ser_error`, which is final, in general catching exceptions by value can have slice issues.

danielaparker:
On second thoughts, may be better not to catch `std::bad_alloc` in the fuzzer. We can follow Intel's tinycbor and have a limit on the data item size, configurable in options.

DavidKorczynski:
Yeah I agree - from a security point of view it would be prefereable if an attacker cant construct a document that causes the library to throw bad_alloc exceptions. I have some more updates that I will PR in a few days with the stuff we have been discussing.

PR #224, opened 2020-04-18
Fuzzers for the encoders

DavidKorczynski:
Cross-referencing: https://github.com/danielaparker/jsoncons/pull/223

I started working on discussion we had above, and integrated the initial encoding fuzzers. Before I went further I just wanted to clarify the specific goal we are going for: the current set up enables us to fuzz the parsers + decoders, and we are looking to expand to also fuzz the encoders, correct? In particular, we are not looking to refine the existing fuzzers as such, but rather to expand with new ones? The goal with the fuzzers is to get as much of the code covered so in my mind it makes sense to expand with the encoders rather than refining the existing ones.

### DIFF
--- a/fuzzers/fuzz_bson_encoder.cpp
+++ b/fuzzers/fuzz_bson_encoder.cpp
@@ -1,16 +1,20 @@
-#include <jsoncons/json_parser.hpp>
 #include <jsoncons/json.hpp>
+#include <jsoncons_ext/bson/bson.hpp>
+#include <jsoncons_ext/bson/bson_reader.hpp>
+#include <catch/catch.hpp>
+#include <sstream>
 
 using namespace jsoncons;
+using namespace jsoncons::bson;
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
 	std::string s(reinterpret_cast<const char*>(data), size);
 	std::istringstream is(s);
+	std::vector<uint8_t> s1;
+	bson_bytes_encoder encoder(s1);
+	bson_reader reader(is, encoder);
 
-	std::string s2;
-	json_string_encoder visitor(s2);
-	json_reader reader(is, visitor);
 	std::error_code ec;
 	reader.read(ec);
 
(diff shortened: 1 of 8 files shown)


danielaparker:
Agreed we shouldn't touch the existing ones, at least until all related oss-fuzz issues have been cleared.

Regarding this new fuzzer, that emits parse events into a `default_json_visitor` (effectively discarding them),

```c++
default_json_visitor visitor;
json_reader reader(is, visitor);
std::error_code ec;
reader.read(ec);
```
I would suggest providing options with a maximum nesting depth, i.e.

```c++
default_json_visitor visitor;

json_options options;
options.max_nesting_depth(std::numeric_limits<int>::max());

json_reader reader(is, visitor, options);
std::error_code ec;
reader.read(ec);
```
By default, the max nesting depth is now 1024 for all parsers. If it exceeds that, we'll get an error code with message "Data item nesting exceeds limit set in options". But we don't know whether that happened because the fuzz input contained a large sequence of '[' characters, or if the parser got into a bad state and kept emitting a `begin_array` event without processing any more input. By setting a maximum nesting depth, we can expect to get a fuzz Timeout failure for the latter case.

Note that we should only set a maximum nesting depth for parsers that are emitting parse events into `default_json_visitor`.

Similar tests can be written for bson, cbor, msgpack, and ubjson using bson_options, cbor_options, msgpack_options and ubjson options. Note that csv doesn't have a maximum nesting depth option because it doesn't nest more than a couple of levels.

Daniel

DavidKorczynski:
Sounds good! In the next few days I will upload fuzzers for the rest of the encoders and also fix the existing one. 
No problem with the work - you have done a lot of work on making this library and it's pretty impressive so I am happy to help doing my small part in making it bulletproof.

DavidKorczynski:
The latest commit should take care of the encoders. I will also update on the OSS-Fuzz end once these are merged in and then I will move on to extend with fuzzers for parsers that allow a huge number of nesting depth levels. Perhaps the last thing we really need for this to be really well covered is adding some corpus files for the fuzzers. Corpus files are simply some files of each data format that the fuzzers can use as a reference for "good" input.

PR #234, opened 2020-05-02
Fixed up build script to include max parsers and new cursor fuzzer

DavidKorczynski:
Could you please verify that the code in `fuzz_json_cursor.cpp` is alright? It runs into an abort quickly, but from my intuition the code looks alright?

### DIFF
--- a/fuzzers/fuzz_json_cursor.cpp
+++ b/fuzzers/fuzz_json_cursor.cpp
@@ -10,11 +10,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
         std::error_code ec;
         json_cursor reader(is, ec);
-        while (reader.done() == 0)
+        while (reader.done() == 0 && ec == 0)
         {
                 const auto& event = cursor.current();
                 std::string s2 = event.get<std::string>();
-                reader.next();
+                reader.next(ec);
         }
 
         return 0;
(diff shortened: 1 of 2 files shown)


danielaparker:
It needs to handle parse errors. So either

```c++
        std::error_code ec;
        json_cursor reader(is, ec);
        while (!reader.done() && !ec)
                reader.next(ec);
```
or
```c++
try
{
        json_cursor reader(is);
        while (!reader.done())
                reader.next();
}
catch (const ser_error&)
{
}
```
Within the loop, you could add
```c++
        const auto& event = cursor.current();
        std::string s = event.get<std::string>();
```
as all event should be convertible into string.

fuzzers/fuzz_json_cursor.cpp
```diff
@@ -8,9 +8,14 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
         std::string s(reinterpret_cast<const char*>(data), size);
         std::istringstream is(s);
 
-        json_cursor reader(is);
+        std::error_code ec;
+        json_cursor reader(is, ec);
         while (reader.done() == 0)
```
danielaparker, line 13: Instead: while (reader.done() == 0 && ec == 0)

fuzzers/fuzz_json_cursor.cpp
```diff
@@ -8,9 +8,14 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
+        {
+                const auto& event = cursor.current();
+                std::string s2 = event.get<std::string>();
                 reader.next();
```
danielaparker, line 17: Instead: reader.next(ec);

PR #273, opened 2020-09-17
Fix float128 feature macro names

### LINKED ISSUE #272: jsoncons defines BOOST_HAS_FLOAT128
jsoncons's compiler_support header is defining BOOST_HAS_FLOAT128. This can lead to some interesting compiler errors when mixing jsoncons and boost, depending on header order:

```
 /usr/include/boost/atomic/detail/type_traits/is_iec559.hpp:37:19: error: 'float128_type' is not a member of 'boost'
  struct is_iec559< boost::float128_type >
```

This is because Boost and jsoncons disagree about when to define this (Boost has some extra checks that have to pass), so if you include some boost headers, then include jsoncons, then include more boost headers, the boost config headers can decide that there isn't a float128 type, while jsoncons decides there is, and then further boost headers think there should be a boost::float128_type when there isn't.

- Compiler: GCC 5.4.0
- Architecture: x64
- Operating system: Ubuntu 16.04

 - [x] Latest release v0.156.0
 - [x] master

amerry:
It was clearly a copy-paste error to define them as BOOST_HAS_FLOAT128,
especially as later code checks for JSONCONS_HAS_FLOAT128.

Fixes #272

### DIFF
--- a/include/jsoncons/config/compiler_support.hpp
+++ b/include/jsoncons/config/compiler_support.hpp
@@ -263,7 +263,7 @@
 #if (defined(linux) || defined(__linux) || defined(__linux__) || defined(__GNU__) || defined(__GLIBC__)) && !defined(_CRAYC)
 #if (__clang_major__ >= 4) && defined(__has_include)
 #if __has_include(<quadmath.h>)
-#  define BOOST_HAS_FLOAT128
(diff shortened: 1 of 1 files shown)


danielaparker:
Indeed! Thanks for spotting that.

PR #279, opened 2020-10-22
variant of json_query() that returns pointers

wbangna:
This makes it possible to change values depending on their previous values.
With json_replace() it is possible to replace a matched value with a new predefined value but it is not possible to rewrite the value depending on the previous value.
Getting the pointers of the matched values allows us to run search-and-replace rules on the old value to generate the new value.

### DIFF
--- a/doc/ref/jsonpath/json_replace.md
+++ b/doc/ref/jsonpath/json_replace.md
@@ -7,9 +7,15 @@ template<class Json, class T>
 void json_replace(Json& root, 
                   const typename Json::string_view_type& path, 
                   T&& new_value)
+
+template<class Json, class Op>
+void json_replace(Json& root, 
+                  const typename Json::string_view_type& path, 
+                  T Op)
 ```
 
-Searches for all values that match a JSONPath expression and replaces them with the specified value
+(1) Searches for all values that match a JSONPath expression and replaces them with the specified value
+(2) Searches for all values that match a JSONPath expression and replaces them with the result of the given function
 
 #### Parameters
 
@@ -23,8 +29,12 @@ Searches for all values that match a JSONPath expression and replaces them with 
     <td>JSONPath expression string</td> 
   </tr>
   <tr>
-    <td>new_value</td>
+    <td>(1) new_value</td>
     <td>The value to use as replacement</td> 
+  </tr>
+  <tr>
+    <td>(2) op</td>
+    <td>Callback function to rewrite values that matched the expression</td> 
   </tr>
 </table>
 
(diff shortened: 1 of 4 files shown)


danielaparker:
I think the goal for this pull request, to allow access to the vector of pointers, is worthwhile. 

But there's an issue with the approach. While typically the pointers in the vector are to values in the original json, there are also cases where the pointers are to instances created during the JSONPath evaluation and stored in the `jsoncons::jsonpath::detail::jsonpath_resources<Json>` object. This happens in [json_query.hpp](https://github.com/danielaparker/jsoncons/blob/master/include/jsoncons_ext/jsonpath/json_query.hpp) wherever `resources.create_temp` is called. An example would be when the query returns the length of a vector, or the result of a function call. The problem is that when your `json_query_ptr` returns, the temporary `jsoncons::jsonpath::detail::jsonpath_resources<Json>` object is gone, and pointers to instances stored there would be left dangling.

wbangna:
Thank you for your quick response!
I wasn't aware of the possibility of temporary nodes.
There is one little problem with your suggestion. `jsonpath_resources` is defined in the `detail` namespace. So I would need to make an alias to make it available in the API.
An other solution would be to use a callback function that gets called for every matching node.

danielaparker:
I like your idea of a callback, because it exposes less implementation detail. Like this

```c++
auto op = [](const json& val) {return std::round(val->as<double>() - 1.0);};

json_replace(booklist, "$.store.book[*].price", op);
```

wbangna:
I tried to implement an overload of json_replace() but I found it pretty hard to distinguish between a callback and all the other possible types. So I wrote a function json_replace_fn() function instead.
I still would prefer to have an overload of json_replace().
My ideas to make this possible would be:
- make an overload with std::function (comes with a runtime cost)
- implementing a trait class that detects callable types
I was a little disappointed that there is no such trait implementation already in the standard library. The only thing I found was std::is_function which does not work for callable objects but only for "real" functions.
I could try to combine the trait std::is_function with a trait that detects whether the operator() is implemented.

danielaparker:
Regarding your second suggestion, implementing a trait class that detects callable types, try

```c++
    template<class FunctionObject, class Arg>
    using
        function_object_t = decltype(std::declval<FunctionObject>()(std::declval<Arg>()));

    template<class FunctionObject, class Arg>
    using
        is_function_object = jsoncons::detail::is_detected<function_object_t, FunctionObject, Arg>;

    template<class Json, class T>
    typename std::enable_if<!is_function_object<T,Json>::value,void>::type
    json_replace(Json& root, const typename Json::string_view_type& path, T&& new_value)
    {
    }

    template<class Json, class Op>
    typename std::enable_if<is_function_object<Op,Json>::value,void>::type
    json_replace(Json& root, const typename Json::string_view_type& path, Op new_value)
    {
    }
```
If that works, `is_function_object` can be moved to `jsoncons/detail/more_type_traits.hpp`, in namespace `jsoncons::detail`.

wbangna:
Wow! Thank you, that worked!
I moved the new trait object to `jsoncons/detail/more_type_traits.hpp` as you whished.

danielaparker:
Thanks for contributing!

PR #298, opened 2020-12-29
add missing copy/move assignment operators to json_object

prince-chrismc:
While trying to use this library I ran into this issue, the copy/move constructor are implemented however the `operator=` where not present or deleted in the top level object. They are however implemented throughout the code back such as:

https://github.com/danielaparker/jsoncons/blob/b9e95d6a0bcb2dc8a821d33f98a530f10cfe1bc0/include/jsoncons/json_container_types.hpp#L416
https://github.com/danielaparker/jsoncons/blob/b9e95d6a0bcb2dc8a821d33f98a530f10cfe1bc0/include/jsoncons/json_container_types.hpp#L426

Perhaps there is a reason but looking at the git blame there was no obvious reason why but feel free to enlighten me 🤓 

I also noticed they are missing from the `json_array` which might be something to considered in the future!

### DIFF
--- a/include/jsoncons/allocator_holder.hpp
+++ b/include/jsoncons/allocator_holder.hpp
@@ -22,7 +22,7 @@ public:
     allocator_holder(allocator_holder&&)  = default;
     allocator_holder& operator=(const allocator_holder&)  = default;
     allocator_holder& operator=(allocator_holder&&)  = default;
-    allocator_holder(const allocator_type&)
+    allocator_holder(const allocator_type& alloc)
         : alloc_(alloc)
         {}
     ~allocator_holder() = default;
--- a/include/jsoncons/allocator_holder.hpp
+++ b/include/jsoncons/allocator_holder.hpp
@@ -18,10 +18,13 @@ private:
     allocator_type alloc_;
 public:
     allocator_holder() = default;
-    allocator_holder(const allocator_type&)  = default;
-    allocator_holder(allocator_type&&)  = default;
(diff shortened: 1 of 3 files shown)


prince-chrismc:
If you are interested in this being added let me know, I'll need to checkout the code apply the changes, the web gui is not sufficient for me

danielaparker:
Can you check the failing test cases?

Thanks,
Daniel

prince-chrismc:
I noticed in quite a few spots you have the deleted functions marked private... that's not required in C++11... future clean up?

Secondly you have a lot of this
https://github.com/danielaparker/jsoncons/blob/5a5037d6523e2464e6528a60ceed660fe66b3c14/include/jsoncons/basic_json.hpp#L530-L537

❓ Why do you implement the copy constructor but delete the copy assignment?

danielaparker:
> I noticed in quite a few spots you have the deleted functions marked private... that's not required in C++11... future clean up?
> 
For classes intended to be used by application developers, they should be public. For internal classes, I don't really care,
one way or the other.

> Secondly you have a lot of this
> https://github.com/danielaparker/jsoncons/blob/5a5037d6523e2464e6528a60ceed660fe66b3c14/include/jsoncons/basic_json.hpp#L530-L537
> 
> ❓ Why do you implement the copy constructor but delete the copy assignment?

The `_storage` classes are for implementation, and the implementation doesn't use copy assignment. For these classes,
we want a minimal interface to discourage accidental misuse.

PR #301, opened 2021-01-02
Fix incorrect custom-specializations.md docs

silvanocerza:
The current documentation [`Custom specialization`](https://github.com/danielaparker/jsoncons/blob/master/doc/ref/json_type_traits/custom-specializations.md#custom-specialization) is wrong and the example showing how to specialize a custom class to support `json_type_traits` fails to compile because the wrong macro is used.


danielaparker:
Thanks!

PR #313, opened 2021-03-31
Get rid of gcc warning -Wnon-virtual-dtor

wbangna:
We get warnings when compiling the jsonpath extension:
`... has accessible non-virtual destructor [-Wnon-virtual-dtor]`

Simply defining a defaulted virtual destructor for the base class `binary_operator` removes the warning.

### DIFF
--- a/include/jsoncons_ext/jsonpath/path_expression.hpp
+++ b/include/jsoncons_ext/jsonpath/path_expression.hpp
@@ -512,7 +512,11 @@ namespace detail {
         virtual JsonReference evaluate(dynamic_resources<Json,JsonReference>&,
                              JsonReference, 
                              JsonReference, 
+
                              std::error_code&) const = 0;
+
+    protected:
+        ~binary_operator() = default;
     };
 
     // Implementations


danielaparker:
Thanks for noting this. Going forward, we'll compile the gcc tests with the `-Wnon-virtual-dtor` compile flags.

I believe though that the `binary_operator` specializations don't require a public virtual destructor, and they're never destroyed through a pointer to a base class. Could you try an alternative approach to silence the warning, by making the base destructor non-virtual but protected? i.e.

```
protected:
~binary_operator() = default;
```
Thanks,
Daniel

wbangna:
Thank you for your suggestion. I guess that it's safe to have a non-virtual destructor since the class is in the namespace `detail`.
I changed the destructor as you suggested and tested it with a small test program.

danielaparker:
> Thank you for your suggestion. I guess that it's safe to have a non-virtual destructor since the class is in the namespace `detail`.
> I changed the destructor as you suggested and tested it with a small test program.

Thanks.

danielaparker:
Thanks, merged (after I initially forgot to merge it before closing it!) The main reason for making it non-virtual protected is that it documents and enforces an internal design choice.

PR #314, opened 2021-03-31
Fix memory leak in jsoncons::jsonpath::node_set

wbangna:
There is a memory leak in node_set when it holds only one node.


danielaparker:
Thanks for contributing!

PR #323, opened 2021-05-26
Add ARCH_INDEPENDENT so the generated cmake config files are really arch independent

### LINKED ISSUE #322: arch independent
Hi,

in the code lines

https://github.com/danielaparker/jsoncons/blob/2ea049807291ff7031006e7f798e3b8b4495e6af/CMakeLists.txt#L61-65

is stated "jsoncons is header-only and does not depend on the architecture."

but the package file generated in those lines still has an architecture check active, which will cause

```
The following configuration files were considered but not accepted:
1> [CMake] 
1> [CMake]     C:/Code/jsoncons/install/lib/cmake/jsoncons/jsonconsConfig.cmake, version: 0.160.0 (64bit)
```
when trying to use the same install for a 32bit build.

only adding `ARCH_INDEPENDENT` to the `write_basic_package_version_file` disables this check.

Flow86:
downside is that it at least requires cmake 3.14

closes #322

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.14)
 
 project(jsoncons CXX)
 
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks for the contribution. We still need to support some environments that require a version of cmake older than 3.14, so I've used VERSION_LESS in CMakeLists.txt to allow that.

PR #336, opened 2021-10-11
introduce inheritance of json_exception from std::exception to comply with sonarqube recommendations and language best practices

pjtallon:
Let json_exception inherit from std::exception

### DIFF
--- a/include/jsoncons/json_exception.hpp
+++ b/include/jsoncons/json_exception.hpp
@@ -10,6 +10,7 @@
 #include <string> // std::string
 #include <sstream> // std::ostringstream
 #include <system_error> // std::error_code
+#include <exception>
 #include <jsoncons/unicode_traits.hpp> // unicode_traits::convert
 #include <jsoncons/config/jsoncons_config.hpp>
 #include <jsoncons/more_type_traits.hpp>
@@ -18,7 +19,7 @@ namespace jsoncons {
 
     // json_exception
 
-    class json_exception
+    class json_exception : public std::exception
     {
     public:
         virtual ~json_exception() noexcept = default;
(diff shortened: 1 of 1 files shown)


danielaparker:
Consider
```
#include <exception>
#include <string>
#include <stdexcept>
#include <iostream>

class json_exception : public std::exception
{
public:
	virtual ~json_exception() noexcept = default;
	virtual const char* what() const noexcept = 0;
};

class key_not_found : public std::out_of_range, public virtual json_exception
{
public:
	key_not_found(const std::string& s)
		: out_of_range(s)
	{
	}

	const char* what() const noexcept
	{
		return std::out_of_range::what();
	}

};

int main()
{
	try
	{
		throw key_not_found("Not found");
	}
	catch (const std::exception& e)
	{
		std::cout << e.what() << "\n";
	}
}
```
clang aborts at runtime, vc++ produces a warning that 'exception': ambiguous (and aborts at runtime.) This can't work unless all derivations of `std::exception` use virtual inheritance, but that's not what `std::out_of_range` does. And almost all `jsoncons` exception classes are derived from `stdexcept` classes.

Note that `jsoncons::json_exception` is similar in role to `boost::exception`, which also does not derive from `std::exception`. boost has exception hierarchies like
```
struct hex_decode_error : virtual boost::exception, virtual std::exception {};
```

danielaparker:
Not sure about `boost::exception`, but the only purpose of `jsoncons::json_exception` is to give users an option to catch all exceptions thrown by jsoncons, should they wish to do so, while at the same time having an exception type hierarchy that derives from the `stdexcept`  classes. 

I think it could be argued that sonarqube is being a little too noisy, that it's helpful to warn about throwing an exception object that's not derived from `std::exception`, but less helpful to warn about catching one.

PR #337, opened 2021-10-13
fixed typos: _cplusplus replaced with __cplusplus

peetonn:
Small fixes for compiler_support.hpp: some places erroneously  had `_cplusplus` instead of `__cplusplus` macro.


danielaparker:
Thanks!

PR #360, opened 2022-05-10
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,9 +9,16 @@ on:
   schedule:
     - cron: '0 19 * * 1'
 
+permissions:
+  contents: read
+
 jobs:
   CodeQL-Build:
 
+    permissions:
+      actions: read  # for github/codeql-action/init to get workflow details
+      contents: read  # for actions/checkout to fetch code
+      security-events: write  # for github/codeql-action/autobuild to send a status report
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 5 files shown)


danielaparker:
Can you look at the four cancelled checks involving Windows / vs2017 builds?

naveensrinivasan:
> Can you look at the four cancelled checks involving Windows / vs2017 builds?

TBH I don't know why? I removed the change for Windows.

naveensrinivasan:
image


Probably this is the reason https://github.com/danielaparker/jsoncons/actions/runs/2302898313 


Even without the Windows change the workflow is failing.

danielaparker:
I see. Could you restore the change for Windows then? I'll accept the pull request, and address the issue with the vs2017 builds separately.

naveensrinivasan:
> I see. Could you restore the change for Windows then? I'll accept the pull request, and address the issue with the vs2017 builds separately.

@danielaparker  I did. The only change is there with the `.bak` file for windows which shouldn't be an issue.

PR #362, opened 2022-05-13
chore: Enable codeql action

naveensrinivasan:
This action runs GitHub's industry-leading semantic code analysis engine,
CodeQL, against a repository's source code to find security vulnerabilities.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,43 @@
+name: "Code Scanning - Action"
+
+on:
+  pull_request:
+
+jobs:
+  CodeQL-Build:
+    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
+    runs-on: ubuntu-latest
+
+    permissions:
+      # required for all workflows
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        # Override language selection by uncommenting this and choosing your languages
+        # with:
+        #   languages: go, javascript, csharp, python, cpp, java
+
+      # Autobuild attempts to build any compiled languages (C/C++, C#, or Java).
+      # If this step fails, then you should remove it and run the build manually (see below).
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2
+
+      # ℹ️ Command-line programs to run using the OS shell.
+      # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+
+      # ✏️ If the Autobuild fails above, remove it and uncomment the following
+      #    three lines and modify them (or add more) to build your code if your
+      #    project uses a compiled language
+
+      #- run: |
+      #     make bootstrap
+      #     make release
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2


danielaparker:
CodeQL has a complaint, which I don't understand, and off hand I don't think it's right. In the code
```cpp
{
        basic_bigint<Allocator> r;
        divide( divisor, *this, r, false );
        return *this;
}

```
CodeQL complains "May return stack-allocated memory from `r`." Function `divide` returns a quotient and a remainder as output parameters, in this case, we don't need the remainder, so we output it to the temporary value `r`, which destructs when the function returns, as it should. At no time does any memory bound to `r` leak into a return value.

naveensrinivasan:
> CodeQL has a complaint, which I don't understand, and off hand I don't see how it could be right. In the code
> 
> ```c++
> basic_bigint& operator/=( const basic_bigint<Allocator>& divisor )
> {
>         basic_bigint<Allocator> r;
>         divide( divisor, *this, r, false );
>         return *this;
> }
> ```
> 
> CodeQL complains "May return stack-allocated memory from `r`." Function `divide` returns a quotient and a remainder as output parameters, in this case, we don't need the remainder, so we output it to the temporary value `r`, which destructs when the function returns, as it should. At no time does any memory bound to `r` leak into a return value, at least I don't see how.

TBH I don't know.

danielaparker:
> > CodeQL has a complaint, which I don't understand, and off hand I don't see how it could be right. In the code
> > ```c++
> > basic_bigint& operator/=( const basic_bigint<Allocator>& divisor )
> > {
> >         basic_bigint<Allocator> r;
> >         divide( divisor, *this, r, false );
> >         return *this;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > CodeQL complains "May return stack-allocated memory from `r`." Function `divide` returns a quotient and a remainder as output parameters, in this case, we don't need the remainder, so we output it to the temporary value `r`, which destructs when the function returns, as it should. At no time does any memory bound to `r` leak into a return value, at least I don't see how.
> 
> TBH I don't know.

I note that it's only reporting one issue. Does CodeQL stop at the first issue it finds, or is that the only issue that it's found? 

Thanks,
Daniel

naveensrinivasan:
> > > CodeQL has a complaint, which I don't understand, and off hand I don't see how it could be right. In the code
> > > ```c++
> > > basic_bigint& operator/=( const basic_bigint<Allocator>& divisor )
> > > {
> > >         basic_bigint<Allocator> r;
> > >         divide( divisor, *this, r, false );
> > >         return *this;
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > CodeQL complains "May return stack-allocated memory from `r`." Function `divide` returns a quotient and a remainder as output parameters, in this case, we don't need the remainder, so we output it to the temporary value `r`, which destructs when the function returns, as it should. At no time does any memory bound to `r` leak into a return value, at least I don't see how.
> > 
> > 
> > TBH I don't know.
> 
> I note that it's only reporting one issue. Does CodeQL stop at the first issue it finds, or is that the only issue that it's found?
> 
> Thanks, Daniel

This is the only issue found.
Also, all the issues would be available in the Security tab that you can review later and dismiss if you think it isn't valid. 

This would be once the PR is merged :)

Thanks

PR #373, opened 2022-06-09
fix include error in g++ 4.8

hongweipeng:

### DIFF
--- a/include/jsoncons/source.hpp
+++ b/include/jsoncons/source.hpp
@@ -37,7 +37,7 @@ namespace jsoncons {
             null_buffer(null_buffer&&) = default;
             null_buffer& operator=(null_buffer&&) = default;
 
-            int_type overflow( int_type ch = traits_type::eof() ) override
+            int_type overflow( int_type ch = std::streambuf::traits_type::eof() ) override
             {
                 return ch;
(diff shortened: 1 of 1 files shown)


danielaparker:
We have test coverage for gcc 4.8, but for 4.8.5. There were a lot of gcc bugs fixed between 4.8.1 and 4.8.5.

Does your code still work with gcc 4.8.2 if you change

```cpp
int_type overflow( int_type ch = std::streambuf::traits_type::eof() ) override
{
    return ch;
}
```

to 
```cpp
 int_type overflow( int_type ch = typename std::basic_streambuf<CharT>::traits_type::eof() ) override
 {
        return ch;
 }
```
?

hongweipeng:
After the change, it can be compiled.

hongweipeng:
Oh sorry, didn't notice that your change is different from the one in the PR. I'll try it tomorrow to see if your changes compile.

hongweipeng:
It can be compile in gcc 4.8.2 when using `typename std::basic_streambuf<CharT>::traits_type::eof()` .

danielaparker:
Okay, if you update your pull request to that, I'll accept it.

PR #386, opened 2022-07-07
Fix and enhance basic_json_diagnostics_visitor

### LINKED ISSUE #385: Problems with basic_json_diagnostics_visitor
Usage of `basic_json_diagnostics_visitor` does not compile in C++11/14 modes due to its `static constexpr CharT visit_foo_name[]` members, which result in "unresolved external" linker errors.

The fix is to add the following below the class declaration:

```c++
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_begin_array_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_end_array_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_begin_object_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_end_object_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_key_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_string_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_byte_string_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_null_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_bool_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_uint64_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_int64_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_half_name[];
    template <class C> constexpr C basic_json_diagnostics_visitor<C>::visit_double_name[];
```

It may be cleaner just to use traditional string literals instead directly in the `std::cout` statements.

With the above fix, when I run the example below, I get the following output, where there is no separation between the label the value:
```
visit_begin_object
visit_keyfoo
visit_uint6442
visit_end_object
```

Even with the above fix, I get the following compile error in Clang 12.0.0 (but not in GCC 9.4.0):

```
/home/emile/Dev/jsoncons/include/jsoncons/json_visitor.hpp:1399:14: error: 'jsoncons::basic_json_diagnostics_visitor<char>::visit_begin_array' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
        bool visit_begin_array(std::size_t length, semantic_tag, const ser_context&, std::error_code&) override
             ^
main.cpp:7:30: note: in instantiation of template class 'jsoncons::basic_json_diagnostics_visitor<char>' requested here
    json_diagnostics_visitor visitor;
                             ^
/home/emile/Dev/jsoncons/include/jsoncons/json_visitor.hpp:954:14: note: hidden overloaded virtual function 'jsoncons::basic_default_json_visitor<char>::visit_begin_array' declared here: different number of parameters (3 vs 4)
        bool visit_begin_array(semantic_tag, const ser_context&, std::error_code& ec) override
             ^
```

The fix is to add the following missing override in `basic_json_diagnostics_visitor`:

```c++
        bool visit_begin_array(semantic_tag, const ser_context&, std::error_code&) override
        {
            std::cout << visit_begin_array_name << std::endl;
            return true;
        }
```

Might I also suggest that a `basic_json_diagnostics_visitor` constructor be added that takes an `ostream` reference to which the diagnostic info will be output. By default, it could be bound to `std::cout` to preserve the current behavior.

```c++
#include <string>
#include <jsoncons/json_parser.hpp>

int main()
{
    using namespace jsoncons;
    json_diagnostics_visitor visitor;
    json_parser parser;
    std::string input(R"({"foo":42})");
    parser.update(input.data(), input.size());
    parser.finish_parse(visitor);
    return 0;
}
```

- Compilers: GCC 4.8.5, GCC 9.4.0, Clang 12.0.0
- Operating system: Linux Mint 20

 - [ ] Latest release 0.168.5
 - [ ] Other release ______
 - [X] master

ecorm:
Fixes #385. I took the liberty of enhancing `basic_json_diagnostics_visitor` as follows:
- Optional indentation according to the nesting level of the visited items
- Ability to specify which output stream to write diagnostic information. By default, it writes to `std::cout` (as before) if `CharT` is `char`.
- Added `:` separator between names and values to avoid confusing output like `visit_uint64123`.

If I went too far with the liberties in enhancing `basic_json_diagnostics_visitor`, then I apologize. Please let me know if that's the case, and I'll roll back the undesired changes.

The test case with UBJSON is to verify that a `basic_json_diagnostics_visitor<char>` works with a parser taking a `bytes_source`. The reason why I picked UBJSON is that it's easy to figure out the input bytes by hand compared to other binary formats.

### DIFF
--- a/include/jsoncons/json_visitor.hpp
+++ b/include/jsoncons/json_visitor.hpp
@@ -7,7 +7,9 @@
 #ifndef JSONCONS_JSON_VISITOR_HPP
 #define JSONCONS_JSON_VISITOR_HPP
 
+#include <iostream>
 #include <string>
+#include <utility>
 #include <jsoncons/json_exception.hpp>
 #include <jsoncons/bigint.hpp>
 #include <jsoncons/ser_context.hpp>
@@ -1362,8 +1364,15 @@ namespace jsoncons {
     template <class CharT>
     class basic_json_diagnostics_visitor : public basic_default_json_visitor<CharT>
     {
+    public:
+        using stream_type = std::basic_ostream<CharT>;
+        using string_type = std::basic_string<CharT>;
+
+    private:
         using supertype = basic_default_json_visitor<CharT>;
         using string_view_type = typename supertype::string_view_type;
+
+        struct enabler {};
 
         static constexpr CharT visit_begin_array_name[] = {'v','i','s','i','t','_','b','e','g','i','n','_','a','r','r','a','y', 0};
         static constexpr CharT visit_end_array_name[] =  {'v','i','s','i','t','_','e','n','d','_','a','r','r','a','y', 0};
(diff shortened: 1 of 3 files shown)


ecorm:
Thanks!

PR #390, opened 2022-08-24
Update catch2 to v2.13.9

### LINKED ISSUE #389: Can not build tests with glibc >= 2.34 since version of catch2 is out of date
**Describe the bug**

# Expected behavior

Tests should build with `glibc >= 2.34`.

# Actual behavior

Tests do not build with `glibc >= 2.34`.

```
jsoncons/third_party/catch/catch.hpp:10791:58: error: call to non-‘constexpr’ function ‘long int sysconf(int)’
10791 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
      |                                                          ^~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/bits/sigstksz.h:24,
                 from /usr/include/signal.h:328,
                 from jsoncons/third_party/catch/catch.hpp:8007,
                 from jsoncons/test/src/main.cpp:2:

```

On a machine with `glibc >= 2.34`:
```
mkdir build
cd build
cmake .. -DJSONCONS_BUILD_TESTS=On
cmake --build .
```

See above.

- Compiler: clang-12
- Architecture (e.g. x86, x64): x64
- Operating system: Ubuntu 22.04.1

 - [ ] Latest release 0.168.5
 - [ ] Other release ______
 - [x] master
 
 # Possible solution
 
 It seems like issue comes from using Catch2. 
 
 https://github.com/catchorg/Catch2/issues/2421
 
This has been resolved some time ago, but we are currently using a pretty old version. Maybe it would be sufficient to just use the latest release of Catch2?

koeberlue:
Fixes #389

### DIFF
--- a/third_party/catch/catch.hpp
+++ b/third_party/catch/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.12.1
- *  Generated: 2020-04-21 19:29:20.964532
+ *  Catch v2.13.9
+ *  Generated: 2022-04-12 22:37:23.260201
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2020 Two Blue Cubes Ltd. All rights reserved.
+ *  Copyright (c) 2022 Two Blue Cubes Ltd. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks!

PR #393, opened 2022-10-24
Fix boost interprocess example

### LINKED ISSUE #392: boost_interprocess examples not working
The library provides two examples that show how to use it with `boost::interprocess`. These examples are not working. A quick look into the code allows to understand that the code is outdated since the policies are not compatible.

After updating the boost policies, it appears that `basic_json` use `std::basic_string` as its `key_type` and does not allow policy customization. It is an issue since `std::basic_string` does not support custom pointer types such as `boost::interprocess::offset_ptr`. 

The examples in `examples_boost` should compile and run properly.

They don't compile.

Compiles the boost_example directory.

- Compiler: gcc 11
- Operating system: AlmaLinux 8

 - [ ] Latest release 0.168.5
 - [ ] Other release ______
 - [x] master

raplonu:
Fixes #392.

All tests ran and passe with success.

Boost examples now compile and produce expected output.

### DIFF
--- a/doc/Tutorials/Custom Allocators.md
+++ b/doc/Tutorials/Custom Allocators.md
@@ -30,17 +30,17 @@ typedef boost::interprocess::allocator<char,
 
 struct boost_sorted_policy : public sorted_policy
 {
-    template <class T, class Allocator>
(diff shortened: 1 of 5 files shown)


danielaparker:
Thanks for contributing!

PR #395, opened 2022-11-03
Add ser_context::end() function

wbangna:
This PR extends the usage of the function `ser_context::position()` to be usable for other elements reported to the visitor (e.g. `start_array`, `end_array` etc.).
It also introduces the function `ser_context::end()` which reports the end position of the currently parsed element.

To test the new functionality use the following test program: https://gist.github.com/wbangna/03bc656af36c50d6479f166f34981682

I did not change the documentation yet as I am not sure if the functionality changes of `ser_context::position()` will be accepted.


danielaparker:
Thanks for contributing. Could you rename `end()` to `end_position()`? Otherwise I'm fine with this request.

wbangna:
Great!
I just renamed the `end()` function and added the documentation for `ser_context`.
Thank you!

Issue #405, opened 2022-12-01
`std::from_chars` parsing fails tests on Windows


cschreib-ibex:
**Describe the bug**

Recently, jsoncons has started using `std::from_char` when available. We are on Windows, with `_MSC_VER == 1934`, which does enable it based on the checks in `compiler_support.hpp`. However, `std::from_chars` seems unable to parse all floating point representations, as demonstrated by tests failing (e.g., `"csv_parser number detection").

I don't know if this is a quirk of the MSVC implementation that affects all versions, or if it is specific to our version of MSVC. But at present the `std::from_char` parsing does not have feature-parity with the classic `strtof`/`strtod`.

**Enumerate the steps to reproduce the bug**

 - Clone jsoncons
 - Build with MSVC (I used VS 17.4.1, `_MSC_VER == 1934`)
 - Run tests

**Include a small, self-contained example if possible**

N/A

**What compiler, architecture, and operating system?**

- Compiler: MSVC 17.4.1, `_MSC_VER == 1934`
- Architecture (e.g. x86, x64) x64
- Operating system: Windows 10

**What jsoncons library version?**

 - [ ] Latest release 0.169.0
 - [ ] Other release ______
 - [x] master

danielaparker:
Thanks. We can revert for now.

Issue #404, opened 2022-11-30
`json_options::inf_to_str` and `neginf_to_str` do not set options correctly for parsing


cschreib-ibex:
**Describe the bug**

There is a logic error in the current implementation of `json_option`: https://github.com/danielaparker/jsoncons/blob/c369979d8cf12c82b643fc71d5261b3d28f7d4d1/include/jsoncons/json_options.hpp#L599-L624

Lines 611 should be `this->enable_str_to_inf_ = enable_inverse;`, and line 620 should be `this->enable_str_to_neginf_ = enable_inverse;`. Otherwise, `enable_inverse` overwrites the `enable_*_to_str_` booleans, and the `enable_str_to_*_` are never set. This was done correctly for `nan`, but not for `inf` and `neginf`.

As a result, it is impossible to parse `inf` and `neginf` strings back into floating point values at the moment (unless the string-to-float parsing functions happen to support the chosen string representation, but this is a happy accident).

**Enumerate the steps to reproduce the bug**

 - Have a Windows machine
 - Clone jsoncons master
 - Compile and run the code below 
 - See "nan" succeeding, and "inf" failing

**Include a small, self-contained example if possible**

Note: the code below "works" by accident as long as jsoncons uses `strtod`/`strtof` to convert strings to floats. These functions naturally support certain representations of infinity and not-a-number. The code fails when jsoncons is configured to use `from_chars` for parsing numbers, which at the moment is only implemented in MSVC for floats.

```c++
#include <fstream>
#include <jsoncons/json.hpp>

int main()
{
    using json = jsoncons::json;
    using json_options = jsoncons::json_options;
    using ostream = std::ofstream;
    using istream = std::ifstream;

    json_options options;
    options.nan_to_str("nan", true);
    options.inf_to_str("+inf", true);
    options.neginf_to_str("-inf", true);

    json j;
    j["nan"] = std::numeric_limits<float>::quiet_NaN();
    j["inf"] = std::numeric_limits<float>::infinity();
    j["neginf"] = -std::numeric_limits<float>::infinity();

    std::string file = "test.json";
    ostream(file) << jsoncons::pretty_print(j, options);

    istream in(file);
    json i = json::parse(in, options);

    try
    {
        std::cout << i["nan"].as<float>() << std::endl;
        std::cout << i["inf"].as<float>() << std::endl;
        std::cout << i["neginf"].as<float>() << std::endl;
    }
    catch (const std::exception& e)
    {
[...]

PR #406, opened 2022-12-01
Multiple fixes for float parsing and boolean pretty print with wjson

### LINKED ISSUE #403: wjson encodes true/false/null with quotes
**Describe the bug**

Consider this simple test code, which just dumps a single boolean field with either `jsoncons::json` or `jsoncons::wjson`:
```c++
#include <fstream>
#include <jsoncons/json.hpp>

template <typename CharT> void test(const char *file) {
  using json = jsoncons::basic_json<CharT>;
  using stream = std::basic_ofstream<CharT>;

  json j;
  if constexpr (std::is_same_v<CharT, char>) {
    j["boolean"] = true;
  } else {
    j[L"boolean"] = true;
  }

  stream(file) << jsoncons::pretty_print(j);
}

int main() {
  test<char>("test_char.json");
  test<wchar_t>("test_wchar.json");
}
```

**What is the expected behavior?**

I expect both created files `test_char.json` and `test_wchar.json` to be identical and contain
```json
{
    "boolean": true
}
```

**What is the actual behavior?**

`test_char.json`:
```json
{
    "boolean": true
}
```

`test_wchar.json`:
```json
{
    "boolean": "true"
}
```

 - Clone jsoncons.
 - Build the example code above (requires C++17).
 - Run the compiled executable.

See above.

- Compiler: gcc 9.4.0 and MSVC 17.4
- Operating system: Linux and Windows

 - [ ] Latest release 0.169.0
 - [ ] Other release ______
 - [x] master

cschreib-ibex:
Fixes #403 #404 #405.

> I changed the CSV parser code to no longer call `substr`; that allowed compilation and tests to pass.

Thanks. Interesting observation about `substr`.

Regarding the remaining failure, I wouldn't expect the test to work with gcc 4.8, as gcc 4.8 `basic_string` had issues with stateful allocators. It's probably enough to suppress the test if 
```
(defined(__GNUC__) && (__GNUC__ == 4) && __GNUC_MINOR__ < 9))
```

### DIFF
--- a/include/jsoncons/config/compiler_support.hpp
+++ b/include/jsoncons/config/compiler_support.hpp
@@ -114,22 +114,6 @@
    #endif
 #endif
 
-#if !defined(JSONCONS_HAS_STD_FROM_CHARS)
-#  if defined(__GNUC__)
-#   if (__GNUC__ >= 11)
-#    if (__cplusplus >= 201703)
-#     if !defined(__MINGW32__)
-#      define JSONCONS_HAS_STD_FROM_CHARS 1
-#     endif // !defined(__MINGW32__)
-#    endif // (__cplusplus >= 201703)
-#   endif // (__GNUC__ >= 11)
-#  endif // defined(__GNUC__)
-#  if defined(_MSC_VER)
-#   if (_MSC_VER >= 1924 && _MSVC_LANG >= 201703)
-#    define JSONCONS_HAS_STD_FROM_CHARS 1
-#   endif // (_MSC_VER >= 1924 && MSVC_LANG >= 201703)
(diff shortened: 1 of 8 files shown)


cschreib-ibex:
On the build failures: I noticed them on my test computer as well. I notice the offending test has a guard against `_MSC_VER == 1900`; I wasn't sure why it was limited to that specific version, but it appears any VS 2022 version is unable to compile this test at the moment. I can change this to `_MSC_VER >= 1900`?

cschreib-ibex:
I changed the CSV parser code to no longer call `substr`; that allowed compilation and tests to pass.

PR #408, opened 2022-12-07
mark jsonpath_expression::evaluate() as const

wbangna:
The Function `jsonpath_expression::evaluate()` should be `const`.


danielaparker:
Thanks for contributing!

PR #431, opened 2023-05-31
Fix macOS build

PhDittmann:
- add `defined()` for `__EXCEPTIONS` and `_MSC_VER` in compiler_support.hpp (see https://github.com/peetonn/ROSBridge2Unreal/commit/cfe4a89451039a4f680fc41c722cf85afd62b131)
- remove shadowing of `using char_type = typename Json::char_type;` in stat_iterator.hpp


danielaparker:
Thanks for contributing!

Issue #432, opened 2023-05-31
Implicit cast from long long to std::size_t in Hash algorithm throws error in win32 configuration


stvchester:
**When building using Visual Studio 2022 through a vcpkg in win32 configuration, I am getting a down casting warning.  As my solution has treat warnings as errors enabled, this prevents compilation.  This warning is new as of 0.170.0 as I rolled back to 0.169.0 and did not have the same issue.**

Compilation without warnings.

Compilation with a down casting warning

I found I was able to resolve the issue by using `std::uintmax_t` instead of `std::size_t` and `long long`.

```cpp
// file: include/jsoncons_ext/jsonpath/expression.hpp
        struct MyHash
        {
            std::uintmax_t operator()(string_type const& s) const noexcept
            {
                const int p = 31;
                const int m = static_cast<int>(1e9) + 9;
                std::uintmax_t hash_value = 0;
                std::uintmax_t p_pow = 1;
                for (char_type c : s) {
                    hash_value = (hash_value + (c - 'a' + 1) * p_pow) % m;
                    p_pow = (p_pow * p) % m;
                }
                return hash_value;
            }
        };
```

**What compiler, architecture, and operating system?**

- Compiler: MSVC 14.35
- Architecture (e.g. x86, x64) x86
- Operating system: Windows 10

**What jsoncons library version?**

 - [x] Latest release 0.170.2

PR #433, opened 2023-05-31
Compilation warning in win32 configuration due to down cast

stvchester:
This is to resolve issue #432 
- Switched hash struct to use std::uintmax_t instead of std::size_t and long long


danielaparker:
Thanks for contributing. Because the master branch has a lot of changes related to propagating stateful allocator support, we'll create a separate patch based on 0.170.2 that includes this change only.

PR #438, opened 2023-07-13
Change signature of dump_pretty

ergonjomeier:
After updating to 0.171.0, our code failed to compile with the following error:

`.../jsoncons/basic_json.hpp:3866:13: error: no matching member function for call to 'dump_pretty' dump_pretty(cont, options, indent, ec);`

The error occurs because on line 3866, dump_pretty is called with 4 parameters, but a dump_pretty function with the corresponding signature does not exists. After looking into the issue (and reading the doc https://github.com/danielaparker/jsoncons/blob/master/doc/ref/corelib/json/dump.md), I think that a dump_pretty function that takes an indent parameter should not exist at all.

I am not entirely sure how to fix this. The docs say that there should be a dump_pretty function that takes a CharContainer and options, with the options having a default parameter. My proposition changes the functions signature to fit the docs.

### DIFF
--- a/include/jsoncons/basic_json.hpp
+++ b/include/jsoncons/basic_json.hpp
@@ -3859,11 +3859,10 @@ namespace jsoncons {
         template <class CharContainer>
         typename std::enable_if<extension_traits::is_back_insertable_char_container<CharContainer>::value>::type
         dump_pretty(CharContainer& cont,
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks for fixing this!

ergonjomeier:
Glad I could help :-)

PR #443, opened 2023-09-04
Fix for shadowed path variable

MonkeybreadSoftware:
Fixes the warning about declaration shadows a local variable for path parameter.


danielaparker:
Thanks for contributing!

MonkeybreadSoftware:
Welcome. I just started using it and we try to have the compiler check as much as possible and fix warnings as much as possible.

PR #444, opened 2023-09-04
Fix for shadowed path variable for JSON Query

MonkeybreadSoftware:
Fixes the warning about declaration shadows a local variable for path parameter.


danielaparker:
Thanks!

PR #447, opened 2023-09-18
Update json_encoder.hpp

MonkeybreadSoftware:
Output semantic_tag::bigdec as number.
We like to use lossless_number when parsing and output these numbers back as numbers and not in quotes.


danielaparker:
Thanks!

PR #456, opened 2023-09-24
Update basic_json.hpp

MonkeybreadSoftware:
Added compare for bigint/bigdec/bigfloat as double or exact text match.

This makes JSON Query work if you use options.lossless_number(true); for parsing JSON.


danielaparker:
Thanks!

PR #481, opened 2024-01-24
make cbegin and cend functions const

amassalha:
Change const iterator functions to const


amassalha:
@danielaparker  welcome
are you plannign a release soon?
I need to use jsoncons from conan with this fix
currently conan has 0.173.2 wihtout the const fix

PR #489, opened 2024-02-14
jsonschema: print validation error location for default error reporter

aldlfkahs:
Default error reporter in jsonschema prints only ```o.message()``` which only shows the reason.  I think it is not enough. Users cannot know where the error occurs. Many other json schema libraries aim for human-comprehensible error messages.
So, how about print little bit more detail? I simply modified by adding ```o.instance_location()``` , referred from custom error reporter example in jsonschema doc.

### DIFF
--- a/include/jsoncons_ext/jsonschema/json_validator.hpp
+++ b/include/jsoncons_ext/jsonschema/json_validator.hpp
@@ -27,7 +27,7 @@ namespace jsonschema {
     {
         void do_error(const validation_output& o) override
         {
-            JSONCONS_THROW(validation_error(o.message()));
+            JSONCONS_THROW(validation_error(o.instance_location() + ": " + o.message()));
         }
     };
 


aldlfkahs:
Additionally, I added ```fail_early``` parameter to ```validate(json, reporter)``` function for custom error reporter as well for personal use.  If you also like this idea, I will make a PR for this too.

danielaparker:
Thanks for contributing. I've merged in this PR, however, we're currently doing a major rewrite of jsonschema on a branch, with the intention of supporting later versions starting with 2019-09, so it's probably better to postpone more jsonschema related PR's on main, until we merge in the new code. We definitely want to improve error messages and reporting, and reporting requirements added in the 2019-09 draft will also be supported in the draft 7 implementation. Any thoughts for further improvements are most welcome, and I'll look at your other suggestion.

danielaparker:
Just FYI, regarding your second suggestion for `fail_early`, since version 0.175.0, the reporter passed to the `json_schema<Json>::validate` functions must return `walk_result::advance` or  `walk_result::abort`, this gives the user control over early exit. (The legacy classes and functions, `make_schema` and `json_validator`, are not affected.)

aldlfkahs:
Great improvement! Thank you for sharing information of new release.

PR #498, opened 2024-04-03
make schema error quotation and punctuation more consistent

mtmorgan:


danielaparker:
@mtmorgan Can you fix ``test/jsonschema/src/validation_report_tests.cpp so that the expected messages match your changes?

PR #510, opened 2024-05-10
fix: avoid over-allocating memory for 8 byte alignment or less

### LINKED ISSUE #509: avoid over-allocatng memory due to alignment constraints
in `heap_string_factory::create` we over-allocate memory to be able to align the pointer withing the allocated block.

However, the default, usual case is that we allocate strings with alignment 8 (= alignof(storage_type).
But all the standard allocators available today in any os already return 8 byte-aligned pointers, so overallocating by 7 bytes is unnecessary. One can think that 7 bytes is not significant but it could push the underlying allocator to select the next size class for the allocation, effectively reserving dozens of bytes more.     

To summarize: standard allocators already return 8 bytes aligned pointers, so we pay penalty for supporting the theoretic edge cases that return non-aligned adresses. 

My suggestion is that for alignments <= 8 we wont allocate exacly the requested size and check if the returned address is aligned. 
And if it hits the unlikely scenario of not being aligned, then deallocate and retry again with the current approach.

romange:
Fixes #509

### DIFF
--- a/include/jsoncons/detail/heap_string.hpp
+++ b/include/jsoncons/detail/heap_string.hpp
@@ -60,6 +60,7 @@ namespace detail {
         pointer p_;
         std::size_t length_;
         uint8_t offset_;
+        uint8_t align_pad_;
 
         ~heap_string() noexcept = default; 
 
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks!

Line 147  

    if (reinterpret_cast<uintptr_t>(ptr) % align == 0)

will fail compilation if ptr is a "fancy pointer". I don't have a test case for that, but you'll see it if you try to compile `boost_examples/interprocess_allocator/shared_memory.cpp`. I think you need to write it as

    q = extension_traits::to_plain_pointer(ptr);
    if (reinterpret_cast<uintptr_t>(q) % align == 0) {

Line 157 

    align_pad = (align-1);

requires a cast to suppress a warning.

romange:
@danielaparker  PTAL

danielaparker:
Thanks for contributing

PR #512, opened 2024-05-21
Fix jsonschema failing example

balleddog:


danielaparker:
Thanks for contributing!

Issue #507, opened 2024-04-24
Compile error on EL7


andysCaplin:
**Describe the bug**

Build fails with lots of notes and some errors

```
In file included from ../../libs/jsoncons/include/jsoncons/basic_json.hpp:30:0,
                 from ../../libs/jsoncons/include/jsoncons/json.hpp:10,
                 from ../src/api_dotnet.cpp:7,
                 from ../src/test_api_dotnet.cpp:9:
../../libs/jsoncons/include/jsoncons/json_object.hpp: In instantiation of ‘void jsoncons::sorted_json_object<KeyT, Json, SequenceContainer>::uninitialized_init(jsoncons::index_key_value<Json>*, std::size_t) [with KeyT = std::basic_string<wchar_t>; Json = jsoncons::basic_json<wchar_t, jsoncons::sorted_policy, std::allocator<char> >; SequenceContainer = std::vector; std::size_t = long unsigned int]’:
../../libs/jsoncons/include/jsoncons/json_decoder.hpp:165:13:   required from ‘bool jsoncons::json_decoder<Json, TempAllocator>::visit_end_object(const jsoncons::ser_context&, std::error_code&) [with Json = jsoncons::basic_json<wchar_t, jsoncons::sorted_policy, std::allocator<char> >; TempAllocator = std::allocator<char>]’
../src/test_api_dotnet.cpp:93:1:   required from here
../../libs/jsoncons/include/jsoncons/json_object.hpp:508:17: error: no matching function for call to ‘std::basic_string<wchar_t>::basic_string(jsoncons::index_key_value<jsoncons::basic_json<wchar_t, jsoncons::sorted_policy, std::allocator<char> > >::key_type&, jsoncons::allocator_holder<std::allocator<char> >::allocator_type)’
                 members_.emplace_back(key_type(first->name,get_allocator()), std::move(first->value));
                 ^
../../libs/jsoncons/include/jsoncons/json_object.hpp:508:17: note: candidates are:
In file included from /usr/include/c++/4.8.2/string:52:0,
                 from ../../libs/jsoncons/include/jsoncons/basic_json.hpp:11,
                 from ../../libs/jsoncons/include/jsoncons/json.hpp:10,
                 from ../src/api_dotnet.cpp:7,
                 from ../src/test_api_dotnet.cpp:9:
[...]

PR #519, opened 2024-06-01
Centos7 build fix

neandreithal:
I've managed to get #507 working with llvm-toolset-7 on CentOS 7. Unfortunately, I can't move to devtoolset-9 as it was done in the original issue.

### DIFF
--- a/include/jsoncons/json_object.hpp
+++ b/include/jsoncons/json_object.hpp
@@ -391,7 +391,7 @@ namespace jsoncons {
             members_.reserve(count);
             for (auto s = first; s != last; ++s)
             {
-                members_.emplace_back(key_type(s->first.c_str(),get_allocator()), s->second);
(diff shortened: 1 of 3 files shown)


danielaparker:
I appreciate the effort!

I'd like to avoid constructing strings with `s.cstr()`, as that would result in a large number of length recalculations, when we already know the length.  

For `key_type`, you should replace
```cpp
key_type(s->first.c_str(), get_allocator())
``` 
with
```cpp
key_type(s->first.data(), s->first.size(), get_allocator())
```
Similarly, you should replace 
```
 keys.emplace(key.c_str(), get_allocator());
```
with
```
 keys.emplace(key.data(), key.size(), get_allocator());
```

neandreithal:
Done, thanks for the note!

neandreithal:
Done, thanks for the feedback!

PR #528, opened 2024-07-18
uri::parse: Fix offset of query_

BenPope:
This was showing up as a heap-buffer-overflow during `operator==`, after calling [this constructor](https://github.com/danielaparker/jsoncons/blob/232e16239fe930e1b5bb13cecd5510abbb6107d5/include/jsoncons/uri.hpp#L109-L113)


danielaparker:
Thanks for contributing!

PR #546, opened 2024-10-25
unexpected behavior for built-in function `join` of  jmespath

Uniplore-X:
Unexpected behavior `join`，when elements are empty strings ahead of the json array. For example: `join(',', @)` 
---
["", "", "a", "","b", "c", ""]
---
current result is:
```
 a,,b,c,
```
but expected:
```
 ,,a,,b,c,
```


danielaparker:
Thanks for this fix!

PR #551, opened 2024-11-10
Fixing example files compilation

2208loki:
- Changed serialization_examples.cpp to use boost to output type _int128, And run chinese_uchar8_t only when __cpp_char8_t is defined.
- Changed CmakeLists to create separate executables for each source file, as it is not possible to compile all the files into single executable because of conflict due to multiple main functions.

With this transaction the example sources except "json_traits_macros_examples.cpp" can be compiled without errors just following instructions in  "jsoncons/examples/build/cmake/README.txt".  

Will try to fix json_traits_macros_examples.cpp in future, but for now removing this file and doing cmake would give separate executables for all the examples.

### DIFF
--- a/examples/build/cmake/CMakeLists.txt
+++ b/examples/build/cmake/CMakeLists.txt
@@ -19,9 +19,16 @@ include_directories (../../include
 
 file(GLOB_RECURSE Example_sources ../../src/*.cpp)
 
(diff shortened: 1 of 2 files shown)


danielaparker:
Thanks! Really appreciate this contribution.

PR #560, opened 2024-11-26
add customer functions for `jmespath`

Uniplore-X:
1. fix a bug of avg function. 
2. add customer functions to enhance the processing capabilities of jmespath extension.(I just tested the example on Ubuntu 22.04.)
3. fix a bug of jmespath test (function expression `"not_null()"` with no args, should result `null` instead of invalid-arity error)

### DIFF
--- a/examples/src/jmespath_customer_functions_examples.cpp
+++ b/examples/src/jmespath_customer_functions_examples.cpp
@@ -1,0 +1,275 @@
+// Copyright 2013-2024 Daniel Parker
+// Distributed under Boost license
+
+#include <chrono>
+#include <thread>
+
+#include <string>
+#include <fstream>
+#include <cassert>
+#include <jsoncons/json.hpp>
+#include <jsoncons_ext/jmespath/jmespath.hpp>
+
+// When adding custom functions, they are generally placed in their own project's source code and namespace.
+namespace myspace
+{
+  using Json = jsoncons::json;
+  using JsonReference = const Json &;
+  using jmespath_errc = jsoncons::jmespath::jmespath_errc;
+#define json_object_arg jsoncons::json_object_arg
+#define json_array_arg jsoncons::json_array_arg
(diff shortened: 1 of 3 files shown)


danielaparker:
Thanks for contributing!

danielaparker:
Just a follow up to your customer functions contribution. I was impressed that you were able to do that given that the essential types you needed were deeply buried as inner classes all inside a detail namespace. I thought your example was quite good too.

Regarding this code and comment 
```cpp
// result->emplace_back(ele); // ?: It may lead to an abnormal exit.
result->emplace_back(*resources.create_json(deep_copy(ele)));
```

the issue is with these lines
```cpp
const auto context = args[0].value();   (*)
// ...
auto ele = expr.evaluate(context, resources, ec2);
```

The first line should have been
```cpp
const auto& context = args[0].value();   (**)
```

`evaluate` expects its reference argument to live through the entire evaluation, but here `context` dies when the function returns.  If written as (**) 

```cpp
result->emplace_back(ele);
```

is fine, as would be (how it would be done internally by evaluate, to minimize copies)
```cpp
const auto& ele = expr.evaluate(context, resources, ec2);
result->emplace_back(json_const_pointer_arg, ele);
```

Looking at the code, I see a lot of `*resources.create_json()` and `std::addressof`. To eliminate most of that, I changed `evaluate` (including function `evaluate`) to return a `const Json*` pointer rather than a `const Json&` reference. I also renamed `null_value()`, `true_value()`, `false_value()` and `create_json()` to `make_null()`, `make_true`, `make_false`, `make_json` and have them consistently return a pointer to a Json value.

I also moved classes `parameter` and `dynamic_resources` out of inner classes and detail namespace, and restructured the code to be as close as possible to custom functions in `jsonpath`. The current version of the example is [here](https://github.com/danielaparker/jsoncons/blob/master/examples/src/jmespath_custom_function_examples.cpp).

Any feedback appreciated.

PR #581, opened 2025-01-12
Remove duplicate type definition(NFC).

c8ef:
```
external/jsoncons~/include/jsoncons/text_source_adaptor.hpp:84:15: error: redefinition of 'value_type'
--
  | using value_type = typename Source::value_type;
  | ^
  | external/jsoncons~/include/jsoncons/text_source_adaptor.hpp:83:15: note: previous definition is here
  | using value_type = typename Source::value_type;
  | ^
  | 1 error generated.

```


danielaparker:
Thanks for contributing!

Issue #589, opened 2025-02-03
JSON as_or accessor


supsm:
**Describe the proposed feature**
It would be nice to have `basic_json::as_or`, which would be like `get_value_or` for `as` instead of `at` (i.e. converting the current basic_json value to the specified type if it matches, or some default value otherwise). This could be useful with `at_or_null`, which would allow the user to perform their own type checking and handling.

**What other libraries (C++ or other) have this feature?**
None that I'm aware of.

**Include a code fragment with sample data that illustrates the use of this feature**
Example 1:
The following code will not throw if the type is incorrect, which may be beneficial in some cases.
(note: this is modified from [the example for get_value_or](https://github.com/danielaparker/jsoncons/blob/ea8d25062f5243035116a21333342ca88e7b6766/doc/ref/corelib/json/get_value_or.md))
(also it seems like the example there is missing some `using`s :)
```c++
#include <jsoncons/json.hpp>

using jsoncons::json;
using jsoncons::json_object_arg;
using jsoncons::json_type;

int parse_category_string(std::string_view sv)
{
    return 0; // placeholder
}

int get_category_id(json j)
{
    json category = j.at_or_null("category");
    if (category.is_string())
    {
        return parse_category_string(category.as<std::string>());
    }
    return category.as_or<int>(100);
}

int main()
{
    json j1(json_object_arg, {{"author","Evelyn Waugh"},{"title","Sword of Honour"}});
    json j2(json_object_arg, {{"author", "J. R. R. Tolkien"}, {"title", "The Lord of the Rings"}, {"category", "fantasy"}});
    json j3(json_object_arg, {{"author", "Robert Louis Stevenson"}, {"title", "Strange Case of Dr Jekyll and Mr Hyde"}, {"category", 123}});

    std::cout << get_category_id(j1) << '\n';
    std::cout << get_category_id(j2) << '\n';
    std::cout << get_category_id(j3) << '\n';
}
```
Output:
```
100
0
123
```

Example 2:
This could also be used to massively simplify cases where invalid values are treated as a default value. In this example, if `key` is not found it will also use the default value (although this can be changed by using `at` instead of `at_or_null`). Provided `j` is a valid json object, this code will not throw.
```c++
json j = ...;
std::string str = j.at_or_null("key").as_or<std::string>("abcdef");
```

danielaparker:
I agree. Would you care to submit a PR?

supsm:
[...]

PR #591, opened 2025-02-09
JSON as_or accessor

supsm:
Implements feature request from #589.

Some caveats/things to consider:
- Numbers will only return the held value (i.e. not the default value) if it matches exactly. Floating-point and integer types are not interchangeable (and string numbers also don't get parsed). It's probably possible to change this, but I'm not sure what the expected behavior should be in this case.
- `get_value_or` sort of has an ambiguous name: it isn't clear that it is getting the value of a member as opposed to the value of the current object (which `as_or` does)
- `as_or` should probably be marked noexcept, but I'm not entirely sure whether something else could throw.

Other things I've noticed:
- in `as` (basic_json.hpp:3288), it is implemented as
```c++
T val = json_type_traits<...>::as(*this);
return val;
```
instead of returning from `json_type_traits::as()` directly. I've copied this behavior for as_or, but I'm wondering if this is something intentional.
- Some of the tests are inconsistent with each other. In particular, the "test proxy get_value_or" test is quite different in structure and format from the "get_value_or test"


danielaparker:
I appreciate the work. But the points you raise do suggest concerns. I need to think about this.

Regarding `get_value_or`, I have to confess that I've never been enthusiastic about the name. nlohmann has a similar function that they call [value](https://json.nlohmann.me/api/basic_json/value/), but that name isn't inspiring either.  At one time I thought of renaming it to `at_as_or`, which I think is suggestive, but my brain said no! In retrospect I think I prefer the lighter notation `get_or`, `get` is commonly used in other programming languages for getting a value by name from a `map`.  The `_or` only applies here if the value isn't found.

 `is<T>()` and `as<T>()` were contributed very early in this project's life, in [PR #3 (2014)](https://github.com/danielaparker/jsoncons/pull/3). `is<T>()` has always been strict, if `is<T>()` passes, it's generally safe to call `as<T>()`. `is<T>()` doesn't guarantee that `as<T>()` won't throw, however, as `as<T>()` may need to allocate memory. `as<T>()` on its own is generally permissive and coerces to the requested type if sensible.

danielaparker:
As you've noted in your comments, as currently implemented,

    j.as<T>()

and 

    j.as_or<T>(<default_value>)

may return different values for some values where `j.as<T>()` doesn't throw. I think that's undesirable (and hard to explain in the documentation.)

An alternative implementation, rather than doing the check first with `j.is<T>()`, would be to attempt `as<T>()` and return **default_value** only if the attempt failed. 

 A simple implementation might be
```cpp
            using T_type_traits = json_type_traits<basic_json,T>;
            JSONCONS_TRY
            {
                T val = T_type_traits::as(*this);
                return val;
            }
            JSONCONS_CATCH(...)
            {
                return static_cast<T>(std::forward<U>(default_value));
            }
```
Relying on `try/catch` isn't ideal, though. We have users that don't want exceptions, and define **JSONCONS_NO_EXCEPTIONS** to turn throws into `std::terminate`.

It would be helpful if  `json_type_traits<basic_json,T>::as()` reported errors with error codes, rather than throwing. I've been thinking about that, and it is doable in a backwards compatible way.

supsm:
I definitely agree that try/catch isn't a great solution. Even for users who have exceptions enabled, the performance impact could be significant (especially since such a function isn't typically expected to be expensive).

Other options including applying `is_number` instead of `is<T>` if `T` is a numeric type, or adding a separate non-strict `is<T>` to `json_type_traits`. These may easier or find use in other contexts, but their performance will probably be slightly worse than an `as` with an error code due to double checking.

Issue #609, opened 2025-04-17
Compatibility with CMake 3.5 has been removed from CMake


SwooshyCueb:
The `CMakeLists.txt` for this project specifies CMake 3.5 as the minimum version supported, which results in all CMake policy settings being set to preserve compatibility with CMake 3.5:
https://github.com/danielaparker/jsoncons/blob/668a1ac182f025f471ebb4ce27f7307a9d92449b/CMakeLists.txt#L1

Unfortunately, compatibility with CMake 3.5 was deprecated in CMake 3.27 and removed in CMake 4.0. (Compatibility with 3.10 is also deprecated as of 3.31.) From [the CMake documentaiton](https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html#policy-version):

> *Changed in version 4.0:* Compatibility with versions of CMake older than 3.5 is removed. Calls to [`cmake_minimum_required(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html#command:cmake_minimum_required) or [`cmake_policy(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_policy.html#version) that do not specify at least 3.5 as their policy version (optionally via `...<max>`) will produce an error in CMake 4.0 and above.
> 
> *Changed in version 3.31:* Compatibility with versions of CMake older than 3.10 is deprecated. Calls to [`cmake_minimum_required(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html#command:cmake_minimum_required) or [`cmake_policy(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_policy.html#version) that do not specify at least 3.10 as their policy version (optionally via `...<max>`) will produce a deprecation warning in CMake 3.31 and above.
> 
> *Changed in version 3.27:* Compatibility with versions of CMake older than 3.5 is deprecated. Calls to [`cmake_minimum_required(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html#command:cmake_minimum_required) or [`cmake_policy(VERSION)`](https://cmake.org/cmake/help/v4.0/command/cmake_policy.html#version) that do not specify at least 3.5 as their policy version (optionally via `...<max>`) will produce a deprecation warning in CMake 3.27 and above.

Attempting to build this project with CMake 4.0 results in the following error:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
[...]

PR #612, opened 2025-04-22
Update CMake policy usage to 3.14

SwooshyCueb:
I've updated the usage of CMake policy to set policies up to CMake 3.14, and made a couple of tweaks to comply with future policies as well.
I didn't go any further than 3.14, as some MSVC-specific policies are included in 3.15, and  I do not have a Windows development environment on which to test them.
Additionally, I was not able to actually test the examples, as it seems the `CMakeLists.txt` tries to import some nonexistent CMake files, but I'm confident that my changes there are appropriate nonetheless.

Solves #609


danielaparker:
Thanks!

Issue #620, opened 2025-06-19
Filter expression


gagikk:
This filter expression `parent.child.*.kind[?@=='b']` is not working for the following data
```
{
  "parent": {
    "child": {
      "a": { "kind": "a" },
      "b": { "kind": "b" },
      "c": { "kind": "b" },
      "d": { "kind": "c" }
    }
  }
}
```
expected to return the array of `["b","b"]` but actually it returns empty array.

jsoncons 1.3.0

danielaparker:
Thanks for the PR!

PR #621, opened 2025-06-19
fix filter expression

gagikk:
This is a fix for the reported issue #620


danielaparker:
Thanks for contributing!

PR #624, opened 2025-07-31
add test cases for jsonpatch where the jsonpointer "" (document root) is involved

ItsBasi:
* "add" specifically states that the entire document gets replaced with "value" (https://datatracker.ietf.org/doc/html/rfc6902#section-4.1)
* "move", "replace", "test" and "copy" do not disallow operating on the root
* "remove" does not say anything about the root either. I would say that it should be an error, since there is no such thing as an "empty" JSON document

I expect that some of these test cases will fail the pipeline. Would need to take a closer look at the bug, but it looks like the code is treating "" (the root) as "/" (the "" key).


dyumin:
Maybe related: #625

PR #628, opened 2025-08-15
Fix error: ‘c’ may be used uninitialized [-Werror=maybe-uninitialized]

DarthGandalf:


danielaparker:
Thanks for contributing!

PR #633, opened 2025-09-10
Fix make_array() call

amerry:
Add missing allocator-aware array and object constructors

The `make_array(array, allocator_type)` overload generated a compile error
when it was instantiated because there was no corresponding `basic_json`
constructor. This adds that missing constructor. A matching `object`
constructor is also added for completeness.


danielaparker:
Thanks for contributing!

PR #636, opened 2025-09-16
Fix more uninitialized warnings

DarthGandalf:


danielaparker:
Thanks for contributing

PR #638, opened 2025-09-24
Fixing Error Handling in catch macro

kemosparc:
Fixing error handling in catch macro


danielaparker:
One comment before we enable the tests, you should remove the parameter `ec` in
```
JSONCONS_CATCH (const ser_error& ec)
```
can be just
```
JSONCONS_CATCH (...)
```
otherwise some of the tests will complain about an unused variable when exceptions are enabled.

danielaparker:
All tests pass. Thanks for contributing.

PR #647, opened 2025-10-27
Better check for JSONCONS_HAS_STD_FROM_CHARS being defned

MonkeybreadSoftware:
Makes it work if JSONCONS_HAS_STD_FROM_CHARS is defined to 0.


danielaparker:
Thanks for contributing!

Issue #648, opened 2025-10-30
jsonschema uri_wrapper strange different in code of two 'append' methods


TheWatcherCode:
"include/jsoncons_ext/jsonschema/common/uri_wrapper.hpp"
Lines: **117** and **130**

In string method usings "uri_.**fragment**()" but in index method usings "uri_.**encoded_fragment**()", but they are should be the same.
It seems for fix it they both should be "uri_.**encoded_fragment**()".

<img width="781" height="641" alt="Image" src="https://github.com/user-attachments/assets/296315d7-7e01-4cb0-8c1e-68fea5fef539" />

**What jsoncons library version?**

- 1.4.3 and master

danielaparker:
It's not really a bug, but I think you're right that they could both use `encoded_fragment()`.

My original rationale was that because the string argument to append could contain characters that might need to be encoded, it was appropriate to append it to the unencoded fragment, and then the whole fragment would be encoded as one in the call to the uri constructor, which does encoding if needed. But it doesn't really matter since any already encoded characters at the beginning of the fragment would not be affected. And it's more efficient to use `encoded_fragment()`.

Feel free if you wish to submit a PR with this change.

PR #649, opened 2025-10-31
issue 648 - fix jsonschema uri_wrapper append encoding

Oplkill:
https://github.com/danielaparker/jsoncons/issues/648


danielaparker:
Thanks for contributing!

PR #658, opened 2025-11-20
Fix incorrect deallocation from PMR when memory was not aligned

pgit:
When testing `jsoncons` and with a custom PMR that tracks allocations and deallocations, I noticed that occasionally more bytes would be freed than were allocated.


danielaparker:
Thanks for this fix.

PR #665, opened 2025-12-02
line_splits is deprecated

MonkeybreadSoftware:
I think you need JSONCONS_NO_DEPRECATED check around this.


danielaparker:
Thanks!

PR #666, opened 2025-12-08
Fixed "always false" typo in double comparison

beryll1um:
I hope I understood the intention behind this correctly. This still shouldn't impact a lot as it's a corner case mitigation.


beryll1um:
@danielaparker, thanks for your work!

beryll1um:
Introduced in 868ec59dac42ed0c99b79873c687230414c6d104

danielaparker:
You're right. Thanks for contributing!

PR #674, opened 2026-01-09
Fix warnings if YYJSON_MSC_VER doesn't exist

JohannAnhofer:
Fixes two more warnings

### DIFF
--- a/include/jsoncons/config/compiler_support.hpp
+++ b/include/jsoncons/config/compiler_support.hpp
@@ -391,30 +391,6 @@
 #   endif
 #endif
 
-/** noinline for compiler */
-#ifndef JSONCONS_NOINLINE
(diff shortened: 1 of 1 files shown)


danielaparker:
Indeed **YYJSON_MSC_VER** doesn't exist, not in jsoncons, it's a macro from the "C" library [yyjson](https://github.com/ibireme/yyjson). I believe it unintentionally ended up in **compiler_support.hpp** as a result of some experimental work inspired by yyjson, but neither macro that references it, **JSONCONS_NOINLINE** or **JSONCONS_ALIGN**, is used in the existing code base. Instead of the fix, can you update the PR to remove those two macros instead?

 Thanks for spotting this.

JohannAnhofer:
ok, removing dead code is always a good idea.

PR #677, opened 2026-02-11
[draft]dirty-draft idea of making json-schema custom reporters being able to do patches

Oplkill:
What do you think of that idea? I find it very useful to allow not only viewing an error in a custom reporter, but also doing something about it via a patch (a local patch of walk schema situation).

In this test example (yes, it's in the wrong place - just a dirty draft testing the idea), I made it possible to patch invalid JSON if a field doesn't exist. I can make that custom reporter JSON Schema validator add default values (if they exist in the schema) when they are required by the schema.

ps. That code compiles and working test checks in clang c++17

### DIFF
--- a/include/jsoncons_ext/jsonschema/common/keyword_validator.hpp
+++ b/include/jsoncons_ext/jsonschema/common/keyword_validator.hpp
@@ -17,6 +17,7 @@
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <optional>
 
 #include <jsoncons/config/jsoncons_config.hpp>
 #include <jsoncons/conv_error.hpp>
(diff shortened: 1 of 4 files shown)


danielaparker:
This won't compile with older versions of the Standard Library. `std::optional<T>` can be replaced with `jsoncons::optional<T>`, and `std::nullopt` with `jsoncons::optional<T>{}`, but there is no substitute for `std::any`.

Oplkill:
Yes, in complete idea there will be 
`template Json 
jsoncons::optional<Json*>`

danielaparker:
It may be easier to explore this under [Discussions](https://github.com/danielaparker/jsoncons/discussions). Would you consider submitting a new discussion that describes the proposed functionality and provides a simple example for how it would work? Thanks!

danielaparker:
Thanks for the idea and proof of concept, along with the example. This is now implemented.

PR #685, opened 2026-03-04
optimize semantic_tag::noesc write_string

Gudulin-Alexey:
This change replaces the per-character loop with a single bulk write: sink_.append(sv.data(), sv.length())

When encoding large JSON strings tagged as semantic_tag::noesc, json_encoder currently writes characters one-by-one via sink_.push_back(*it). This is a hot path for payloads dominated by large base64 fields and causes a noticeable regression versus previous versions.
This gives significant improvement on noesc-heavy DOM->string


danielaparker:
Thanks for contributing!

PR #687, opened 2026-03-12
jmespath: allow rhs_expression in a keyvalue - expression

BartelNieuwenhuyse:
I  tried the following jmespath expression and noticed that jsoncons_ext jmes_path could not compile the expression:

`{value: foo.bar == 'baz'}`

because it expected '}' or ',' after the lhs_expression value part of a "key: value" and found a comparator ==, i.e. the start of a rhs_expression.

The jmespath grammar allows any expression as value in a multi select hash, so also those consisting of lhs and rhs part.


danielaparker:
Thanks for contributing!

PR #688, opened 2026-03-17
JMESPath functions floor and ceil return integer

BartelNieuwenhuyse:
JMESPath function floor and ceil return a json float number, while the specs and the examples in it suggest that an integer should be returned.

Expected:
``floor(`3.7`) -> 3``
Actual:
``floor(`3.7`) -> 3.0``

### DIFF
--- a/include/jsoncons_ext/jmespath/jmespath.hpp
+++ b/include/jsoncons_ext/jmespath/jmespath.hpp
@@ -1240,8 +1240,8 @@ namespace detail {
     template <typename Json>
     class jmespath_evaluator 
     {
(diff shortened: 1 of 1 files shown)


danielaparker:
The cases that fail are failing because a double doesn't have enough precision to represent `std::numeric_limits<int64_t>::max()` exactly. The upper bound for preserving precision exactly is `std::pow(2, std::numeric_limits<double>::digits)`. Numbers above that limit should remain in double representation. double values above that bound can't be sensibly cast to int types.

danielaparker:
I think this is fine, thanks!

PR #689, opened 2026-03-21
JMESPath functions floor and ceil return integer

BartelNieuwenhuyse:
I added some edge case tests and noticed that my code in pull request [688](https://github.com/danielaparker/jsoncons/pull/688) did cause incorrect results due to overflow:

``ceil(`9223372036854775807`)`` returned -9223372036854775808 instead of +9223372036854775807

So I coded different bounds to check the cast from double to int64.

Sorry for the inconvenience.


danielaparker:
Thanks!

Issue #690, opened 2026-03-25
JMESPath let expression with operator in variable binding


BartelNieuwenhuyse:
JMESPath expressions with an operator in the expression part of a variable binding (expression between 'let $var=' and 'in') are not parsed correctly.

given:
`{"foo": false, "bar": true}`

search expression:
`let $var = foo || bar in $var`

**expected:**
yields result `true`

**actual:**
gives error during evaluation: `assertion 'stack.size() == 1' failed`

**Include a small, self-contained example if possible**
```cpp
TEST_CASE("jmespath let with operator binding expression")
{
    SECTION("Test binary or operator")
    {
        auto doc = jsoncons::json::parse(R"({"foo": false, "bar": true})");
        auto expected = jsoncons::json::parse(R"(true)");

        std::string query = R"(let $op = foo || bar in $op)";
        auto expr = jmespath::make_expression<jsoncons::json>(query);

        jsoncons::json result = expr.evaluate(doc);
        CHECK(expected == result);
    }
}
```

danielaparker:
Thanks! I'm guessing there's an issue with the handling of **expr_state::expect_in_or_comma**. Will need to be investigated.

BartelNieuwenhuyse:
I can 'work around' it using parenthesis:

search expression:
`let $var = (foo || bar) in $var`

gives the expected result.

I'll have a look at the parenthesis logic '(' ... ')' and see if I can make a fix that applies the same logic to '=' ... 'in'

BartelNieuwenhuyse:
Thank you for merging the related pull request #691, but you @danielaparker stated "expression_context serves no other purpose, and can be removed." Is removing expression_context something that will done, or should I open a new pull request for it?

danielaparker:
Feel free to submit a PR to remove it.

BartelNieuwenhuyse:
PR #693

PR #691, opened 2026-03-27
JMESPath let expression with operator in variable binding

BartelNieuwenhuyse:
Fix for issue #690

> JMESPath expressions with an operator in the expression part of a variable binding (expression between 'let $var=' and 'in') are not parsed correctly.
**given:**
{"foo": false, "bar": true}
**search expression:**
let $var = foo || bar in $var
**expected:**
yields result true
**actual:**
gives error during evaluation: assertion 'stack.size() == 1' failed

I've added an implicit lparen '(' when encountering the '=' of the let expression, such that the operator stack can be rewind when hitting the ',' or 'in' of the let-expression.

I've stored the name of the variable in a variable token when encountering the '=', rather than storing it in the expression_context. And use this variable token to mark the start of the binding expression, rather than storing the end_index in the expression_context. I'm not sure if after this change, expression_context has any other use as it seems to be an empty class now.

When hitting the ',' or 'in' of the let-expression the variable token and the tokens of the expression, including the operator stack up until the implicit lparen '(' are moved into the variable_expression, which on evaluation will evaluate the tokens, store the result in the context and return the original document at the start of the let expression.

Moved this logic to push_token, rather than (duplicating it) in the compile function

### DIFF
--- a/include/jsoncons_ext/jmespath/jmespath.hpp
+++ b/include/jsoncons_ext/jmespath/jmespath.hpp
@@ -3142,8 +3142,8 @@ namespace detail {
             token<Json> variable_;
             std::vector<token<Json>> tokens_;
         public:
-            variable_expression(std::vector<token<Json>>&& tokens, token<Json>&& variable)
-                : tokens_(std::move(tokens)), variable_(std::move(variable))
+            variable_expression(token<Json>&& variable, std::vector<token<Json>>&& tokens)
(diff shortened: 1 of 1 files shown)


danielaparker:
I figured this was coming :-) I'll review it when I have a chance. But I think the approach is basically right.

danielaparker:
Thanks. Looks good. You're right, `expression_context` serves no other purpose, and can be removed.

PR #694, opened 2026-04-10
JMESPath expression with operator

BartelNieuwenhuyse:
I think another rewind of the operator stack is missing in case of an expression

given:
`[ { "foo": "bar" }, { "foo": "baz" } ]`

search expression:
`map(&foo == 'bar' && 'yes' || 'no', @)`

expected:
`[ "yes", "no" ]`

actual:
gives error during evaluation: assertion 'stack.size() >= 2' failed

I can add parens, and then it works:
search expression:
`map(&(foo == 'bar' && 'yes' || 'no'), @)`


danielaparker:
Thanks!

PR #698, opened 2026-04-21
JMESPath short circuit or and operator

BartelNieuwenhuyse:
Code change to short circuit the logical and `&&` and or `||` operators:

- `lhs && rhs`: if lhs is false, return it without evaluating rhs
- `lhs || rhs`: if lhs is true, return it without evaluating rhs

For this change, I made an operator_base as base class for the unary (not `!`) and binary (comparators `==`, `<`, etc.) operators and a new logical (and `&&` and or `||`) operator. operator_base has a single interface: operator_base::evaluate takes a reference val and a vector of expressions, which contains 1 expression for the unary and 2 expressions [lhs, rhs] for the binary and logical operators.

To hold this vector, I introduced the operator_expression which references an operator_base and is the owner of a vector of expressions. When evaluating, operator_expression::evaluate forwards the value it gets plus the expressions it owns to the operator_base.

The compilation happens as follows:

- tokens are compiled and placed on the output_stack
- when an operator is encountered, it collects its left hand side tokens by sweeping back over the output_stack and collecting everything belonging to its left hand side. It is then placed on the operator_stack
- when it is popped from the operator_stack, it again sweeps over the output_stack to collect everything belonging to its right hand side. And then placed on the output_stack.


danielaparker:
Nice! I've merged this PR into the branch jmespath_binary_op where I'll have the opportunity to study it more carefully and make sure there's sufficient test coverage for the different possibilities on the left and right sides.

BartelNieuwenhuyse:
Thank you!

PR #700, opened 2026-04-28
Fix bad multiplier when writing some BSON datetimes

sandorzm:
Fix that `basic_bson_encoder::visit_uint64` converted `semantic_tag::epoch_nano`-tagged integers to seconds, not milliseconds, before serializing as a BSON datetime. The resulting datetime would be much closer to 1970-01-01 (the Unix epoch) than it should. This commit copies the correct behavior from `visit_int64` to `visit_uint64`.


danielaparker:
Thanks for contributing!

PR #701, opened 2026-04-30
Don't write extra value after some CBOR epoch timestamps

sandorzm:
Fix that `epoch_milli`- and `epoch_nano`-tagged integers serialized to too many values in CBOR:
1. a CBOR semantic tag of 1 (expected),
2. a float64 number of seconds since the epoch (expected), and then
3. the integer number of milliseconds or nanoseconds (bad).

Only the semantic tag and float64 should be written; the trailing integer could wrongly be interpreted as the next key in a map or value in an array.

Thanks for your work on the library and your quick response to my last pull request.


danielaparker:
Thanks for the fix!

Issue #702, opened 2026-05-01
JMESPath - Merge function can't merge when receiving a json_const_ref


BartelNieuwenhuyse:
JMESPath expressions in which the merge function is used may give the following error if the first argument of the merge function is of storage kind json_const_ref. This because the insert_or_assign method is not implemented for storage kind json_const_ref:

`Attempting to access a member of a value that is not an object`

Various JMESPath functions produce a json_const_ref, so if one of these is used as input to a merge function, the JMESPath expression doesn't work:
- map
- projections (object, list, flatten, slice, filter)
- multi select list and hash

**Enumerate the steps to reproduce the bug**
given: 
```json
{ "foz": "baz" }
```

expression with, for example, a multi select hash:
`{ object: @ }.merge(object, { foo: 'bar' })`

expected:
```json
{ "foo": "bar", "foz": "baz" }
```

actual:
error `Attempting to access a member of a value that is not an object: 'foo'`

**What jsoncons library version?**

 - [ ] Latest release 1.7.0
 - [ ] Other release ______
 - [x] master

danielaparker:
`json_const_ref` is an optimization that allows json values to hold references to already existing json values and so minimize expensive copies while evaluating a chain of expressions. At the end of the chain, before returning a value to the user, a deep copy must be made to make a self contained value.

In the merge function, the issue is probably that 
```cpp
    auto result = context.create_json(arg0);
```
needs to be
```cpp
    auto result = context.create_json(deep_copy(arg0));
 ```

I believe the jmespath extension is the only part of this library that uses `json_const_ref`.

BartelNieuwenhuyse:
Thank you for the clarification. I think a shallow copy will be enough for the merge function and retain some of the benefits of using refs instead of copies during the jmes evaluation. 

I'll submit a pull request once I have it ready.

PR #703, opened 2026-05-06
JMESPath - Merge function can't merge when receiving a json_const_ref

BartelNieuwenhuyse:
Do a shallow copy of the first argument to the merge function: if it is a json_const_ref no modification of the referenced json is allowed.

PR for issue: #702


danielaparker:
Thanks.

PR #705, opened 2026-05-12
JMESPath - Sort and sort_by functions can't sort when receiving a json_const_ref

BartelNieuwenhuyse:
Do a shallow copy of the argument to the sort and sort_by function: if it is a json_const_ref no modification of the referenced json is allowed.

Same problem as #703 but then in the sort and sort_by functions

### DIFF
--- a/include/jsoncons_ext/jmespath/jmespath.hpp
+++ b/include/jsoncons_ext/jmespath/jmespath.hpp
@@ -2121,6 +2121,9 @@ namespace detail {
                     return context.null_value();
                 }
 
+                auto result = context.create_json(json_array_arg);
+                result->reserve(arg0.size());
+                result->push_back(arg0.at(0));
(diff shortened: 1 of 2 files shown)


danielaparker:
Thanks. I think I may need to have another look at `json_const_ref`, it appears to be very easy to misuse. I've thought about introducing a wrapper class, `json_view`, but there are difficulties there too.

BartelNieuwenhuyse:
I think it would be clearer if storage_kind json_const_ref behaves more like a const&

given `a` has storage_kind json_const_ref

Then this would be ok:
`const Json& b = a;`

This could make an implicit shallow copy (like I did manually):
`Json b = a;`

And this would be a compile time error:
`Json& b = a;`

danielaparker:
Moving this to [Discussions ](https://github.com/danielaparker/jsoncons/discussions/706#discussion-10072792)

PR #716, opened 2026-06-05
fix bigdec mantissa and exponent parsing

ky438:
this fixes various CBOR bigdec cases where a 0 would lead to unintended throwing or octal interpretation e.g. "0.8" would throw, "0.215" would be misparsed as "0.141"

also adds support for +, leading dot and fraction followed by exponent, e.g. "+2.5" or ".6" or "-1.5e6"


danielaparker:
Thanks!

PR #717, opened 2026-06-09
fix misplaced #endif

syntax-case:
1.8.0 doesn't work under JSONCONS_NO_DEPRECATED, due to the end of a class getting removed.
The problematic directive was introduced in 5d2f83d.

### DIFF
--- a/include/jsoncons/basic_json.hpp
+++ b/include/jsoncons/basic_json.hpp
@@ -4992,8 +4992,8 @@ namespace jsoncons {
                     return other;
             }
         }
-    };
 #endif
+    };
 
     // operator==
 


danielaparker:
Thanks! I've deleted 1.8.0, and will release a patched version once the tests run. I'll need to check why this wasn't caught in tests, we have workflows with JSONCONS_NO_DEPRECATED.

syntax-case:
I think  `-DCXXFLAGS="-DJSONCONS_NO_DEPRECATED"` should be `-DCMAKE_CXX_FLAGS="-DJSONCONS_NO_DEPRECATED"` in the gcc_build_no_deprecated check.

PR #720, opened 2026-07-07
Add zero-copy CBOR views

ky438:
- add source-level exact span reads and use them for definite-length CBOR text/byte strings
- add raw CBOR view utilities for item spans, skipping, etc
- add coverage for zero-copy cursor views, stream straddling, raw item spans, collation, and invalid view input

built with gpt5.5-xhigh

### DIFF
--- a/include/jsoncons/sink.hpp
+++ b/include/jsoncons/sink.hpp
@@ -347,7 +347,6 @@ namespace jsoncons {
         static auto append_impl(C& c, const uint8_t* s, std::size_t length, long)
             -> decltype(c.insert(c.end(), s, s + length), void())
         {
-            reserve(c, length, 0);
             c.insert(c.end(), s, s + length);
         }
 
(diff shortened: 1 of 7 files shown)


ky438:
Follow-up:

- add bulk append support for bytes_sink and use it for CBOR text/byte payload writes
- expose cbor::view::map_entry_view and cbor::view::map_entries() for zero-copy map traversal
- make cbor::view::compare() compare indefinite/chunked text and byte strings semantically
- add tests for map entry spans and semantic chunked string comparison

ky438:
Follow-up fix for the CIFuzz failure on fuzz_cbor:

- bound stream_source::read_span() growth so malformed huge string lengths cannot trigger an unbounded allocation
- fall back to the existing copy path when a safe span is unavailable
- add a regression test for the fuzzed test case
- avoid exact reserve before bytes_sink insert-based appends
- share CBOR text/byte length header encoding through one helper

danielaparker:
Thanks! This contribution is very much appreciated.

PR #721, opened 2026-07-08
harden cbor_view and add true RFC 8949 ordering

ky438:
- RFC 8949 ordering
- Bound view recursion depth
- Cap untrusted-length allocations
- Deduplicate encoder/parser length writes

also significantly smaller after some manual refactoring


danielaparker:
In source.hpp, these checks aren't right,
```cpp
            if (length > default_max_buffer_size)
            {
                return;
            }

            if (length > default_max_buffer_size)
            {
                return span<const value_type>();
            }
```
`default_max_buffer_size` is the default buffer size, but it can be overridden in the constructor of `xxx_source`. The actual buffer size is usually stored in a data member `buffer_size_`.  We have some test cases where we set the buffer size to as low as 4 bytes to test incremental parsing.

danielaparker:
I've moved your contribution to a develop branch, where I can study it more carefully, and reverted master back to an earlier commit.

PR #722, opened 2026-07-09
expose read_head in cbor::view namespace

ky438:
symmetric with its peers

### DIFF
--- a/include/jsoncons_ext/cbor/cbor_view.hpp
+++ b/include/jsoncons_ext/cbor/cbor_view.hpp
@@ -356,6 +356,11 @@ namespace view {
         }
     };
 
+    inline bool read_head(const uint8_t*& p, const uint8_t* end, item_head& head, std::error_code& ec)
+    {
+        return detail_view::read_head(p, end, head, ec);
+    }
+
     inline bool skip_item(const uint8_t*& p, const uint8_t* end, std::error_code& ec,
         int max_nesting_depth = default_max_nesting_depth)
     {


danielaparker:
Merged into develop

ky438:
> Merged into develop

Great idea, thanks!

danielaparker:
All github workflow tests pass on develop, however, on my desktop, Visual Studio in Debug Mode gives a SIGSEGV - Stack overflow for cbor_cursor_tests.cpp cbor raw view utilities deeply nested items are bounded  test. Release Mode is okay.

This a big PR. It may be too big :-) I'd like to get your contribution for using string_view rather than copy when possible into master before taking in cbor_view. All binary formats will benefit from that. I'm considering merging those changes from develop into master manually. That needs to go along with more tests for source, and source_reader can be improved. I'd like to get those parts thoroughly tested and exposed to Google fuzz before cbor_view.

There are some issues that need to be addressed, in source.hpp, these checks aren't right,

            if (length > default_max_buffer_size)
            {
                return;
            }

            if (length > default_max_buffer_size)
            {
                return span<const value_type>();
            }

default_max_buffer_size is the default buffer size, but it can be overridden in the constructor of xxx_source. The actual buffer size is usually stored in a data member buffer_size_.

danielaparker:
I've created a new branch develop_no_view, which is develop but with cbor_view.hpp (and associated tests in cbor_cursor_tests.cpp) removed. This will preserve the history of your contributions, but make it a smaller amount of changes which are easier to test in isolation. I plan to add some tests, improve or replace source_reader, extend these changes to the other binary formats, address any issues, and then move it to master.

We can then revisit cbor_view as a separate item.

ky438:
Thanks. These are more great ideas and much appreciated. I will do what I can to help.

PR #723, opened 2026-07-14
ky/source buffer guards

ky438:
- **Expose read_head in cbor::view namespace**
- **Compare buffer guards against configured buffer size**

### DIFF
--- a/include/jsoncons/source.hpp
+++ b/include/jsoncons/source.hpp
@@ -417,38 +417,13 @@ namespace jsoncons {
             }
         }
     private:
-        void reserve_buffer(std::size_t capacity)
-        {
-            if (capacity <= buffer_size_)
-            {
-                return;
-            }
-
-            value_type* new_buffer = std::allocator_traits<char_allocator_type>::allocate(alloc_, capacity);
-            if (length_ > 0)
-            {
-                std::memcpy(new_buffer, data_, sizeof(value_type)*length_);
-            }
-            if (buffer_)
-            {
-                std::allocator_traits<char_allocator_type>::deallocate(alloc_, buffer_, buffer_size_);
-            }
-            buffer_ = new_buffer;
-            buffer_size_ = capacity;
-            data_ = buffer_;
-        }
-
         void make_available(std::size_t length)
         {
             if (length > buffer_size_)
             {
-                if (length > default_max_buffer_size)
-                {
-                    return;
-                }
-                reserve_buffer(length);
+                return;
             }
-            else if (length_ > 0 && data_ != buffer_)
+            if (length_ > 0 && data_ != buffer_)
             {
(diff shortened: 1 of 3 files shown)


ky438:
Sorry, I wasn't sure whether this should target master or develop - or exist at all.

ky438:
sorry for the CI noise, this seems to be a genuine issue with g++ versions prior to 13.

danielaparker:
I created a branch from develop with your earlier PR changes, but omitting **cbor_view.hpp** and related test cases in **cbor_cursor_tests.cpp**. This preserves the history of your contributions excluding cbor view, which I would like to keep separate. 

I made a few changes. I added a buffer parameter to the source `read_span` function, so if `read_span` needs to return more bytes than will fit in the source internal buffer, it will make use of the passed buffer. That allows us to replace

```cpp
        auto data = source_.read_span(length);
        if (data.size() == length)
        {
            string_view_type sv(reinterpret_cast<const char_type*>(data.data()), data.size());
           // Store stringref if needed
            return sv;
        }

        text_buffer_.clear();
        if (source_reader<Source>::read(source_, text_buffer_, length) != length)
            // handle error
        string_view_type sv(text_buffer_.data(), text_buffer_.size());
       // Store stringref if needed
       return sv;
```
with
```cpp
        auto data = source_.read_span(length, text_buffer_);
        if (data.size() != length)
            // handle error
        string_view_type sv(reinterpret_cast<const char_type*>(data.data()), data.size());
       // Store stringref if needed
       return sv;
```  
I also removed the code to change the internal buffer size from `stream_source::read_span`. It's fundamental to the design that we read data from streams in fixed length chunks. jsoncons is integrated with Google fuzz, and Google fuzz is constantly bombarding the CBOR parser with CBOR messages that contain string lengths that vastly exceed the size of the message payload, to simulate an attack scenario. By reading streams of bytes in chunks, io will detect an end-of-file before experiencing an allocation failure. It's different when our source is a container of bytes, in which case the chunk is the entire container. 

I've moved these changes into master. I expect them to result in significantly fewer allocations, so a big thanks! In addition I've incorporated these changes into the bson, msgpack, and ubjson parsers so they will all benefit from them.

Regarding cbor view, I regard that as experimental and would like to keep it separate, so not combined with any other changes. Could you create a new PR based on current master, includes **cbor_view.hpp**, and separates the cbor view tests that are in **cbor_cursor_tests.cpp** into a separate file, **cbor_view_tests.cpp**? When you've submitted that PR, I'll accept it into a new branch `cbor_view`. Thanks.

ky438:
Thanks for your comments, and from making these improvements! I agree, allocations should be substantially reduced. 

I am going to close this PR, get to work, and submit a new PR along the lines you have described. I hope to have that done in under a week.

PR #724, opened 2026-07-14
Fix CBOR parser defects found by fuzzing

ky438:


ky438:
Once (if) this is merged, I'll prepare another PR that adds the fuzzers - not doing here to avoid breaking CI

danielaparker:
Thanks. Can you submit a new PR with only these changes that targets master rather than develop? Thanks.

ky438:
Certainly. Closing this now, will submit the new PR when able.

Issue #725, opened 2026-07-15
Bad character in corelib/json/operator_at.md


gri6507:
When jsoncons is packaged as s Debian package, it is possible to run Debian's `lintian` utility against the resulting `libjsoncons-doc` package.

# Expected behavior
No Lintian warnings

# Actual behavior
W: libjsoncons-doc: national-encoding [usr/share/doc/libjsoncons-dev/doc/ref/corelib/json/operator_at.md]

- Compiler: irrelevant
- Architecture: irrelevant
- Operating system: irrelevant, but visible on any Debian based OS

**What jsoncons library version?**
 - [x ] Latest release 1.8.1

gri6507:
Created https://github.com/danielaparker/jsoncons/pull/726

PR #726, opened 2026-07-15
Issue #725 - Fixed non UTF-8 character in doc/ref/corelib/json/operat…

gri6507:
PROBLEM STATEMENT:
`doc/ref/corelib/json/operator_at.md` has a non UTF-8 character (0x92) instead of a UTF-8 compatible apostraphe. This results in locale-aware tools like Debian's `lintian` to report

```
W: libjsoncons-doc: national-encoding [usr/share/doc/libjsoncons-dev/doc/ref/corelib/json/operator_at.md]
```

SOLUTION:
Replaced 0x92 character with latin1 apostraphe `'`

TESTING:
Reran lintian on updated file, which no longer shows the warning.

### DIFF
--- a/doc/ref/corelib/json/operator_at.md
+++ b/doc/ref/corelib/json/operator_at.md
@@ -77,7 +77,7 @@ Output:
     }
 }
 ```
-Note that if `file_export["File Format Options"]` doesn’t exist, the statement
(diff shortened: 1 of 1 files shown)


danielaparker:
Thanks for the fix!

PR #730, opened 2026-07-22
Recast cbor_view as checked encoded-item layer

ky438:
OK - I had a long hard look and wrangled the robot into something I am much happier with.

A view fuzzer is also here, and parser fixes are in https://github.com/danielaparker/jsoncons/pull/731


ky438:
Thanks, @danielaparker ! Dogfooding this further, I can make some further improvements, which I'll make in another PR on the cbor_view branch soon.

PR #731, opened 2026-07-22
Fix CBOR parser well-formedness defects

ky438:
truncated-integer and split-UTF-8 inputs that used to decode now (correctly) error, which I believe is the last disagreement between parser and view with regard to validity


danielaparker:
Thanks!

PR #739, opened 2026-09-07
Constrain and document supported JSON Schema types

### LINKED ISSUE #734: `json_schema` fails to compile with `wjson`
**Describe the bug**

`jsonschema::make_json_schema` seemingly accepts any `Json` type, but fails to compile with a `wjson` at a random location in the code:
```cpp
C:\Users\Ron\Code\jsoncons\include\jsoncons_ext\jsonschema\json_schema_factory.hpp(43): error C2665: 'jsoncons::basic_json<wchar_t,jsoncons::sorted_policy,std::allocator<char>>::find': no overloaded function could convert all the argument types
```

The error is at this line:
https://github.com/danielaparker/jsoncons/blob/01218c7ee8457ab5542051b14150ea12c05fdeaa/include/jsoncons_ext/jsonschema/json_schema_factory.hpp#L43

Replacing `"$schema"` with `JSONCONS_CSTRING_CONSTANT(Json::char_type, "$schema")` works, but there are many other places where adjustments need to be made in order for the code to be `CharT`-agnostic. Fixing that line makes compilation fail elsewhere.

I would expect either full support for all `Json` types, or an explicit `enable_if` for `json`/`ojson` only.

Try to create a `json_schema<wjson>` object.

```cpp
using jsoncons::wjson;
namespace jsonschema = jsoncons::jsonschema;

wjson schema = wjson::parse(LR"(
{
    "properties": {
        "foo": {
            "type": "string"
        }
    }
}
)");

jsonschema::json_schema<wjson> compiled = jsonschema::make_json_schema(schema);
```

- Compiler: MSVC
- Operating system: Windows

 - [x] Latest release 1.8.1
 - [ ] Other release ______
 - [x] master

Alb3e3:
Fixes #734.

`make_json_schema(wjson)` currently reaches narrow-string operations in the schema factories before failing to compile. Constrain all four overloads to `char` JSON types with default-constructible allocators, and add explicit diagnostics for direct `json_schema` instantiation. Document the requirements on the extension and both API pages.

The allocator check also inspects the underlying allocators of `std::scoped_allocator_adaptor`: its declared default constructor can appear usable to `std::is_default_constructible` even when an underlying allocator deletes its default constructor. Default-constructible allocators, including PMR allocators, remain supported.

Tests cover overload selection with default and explicit options, rejection of wide-character and unsupported outer/inner allocator types, and validation through all four overloads with `json`, `ojson`, and their PMR variants. The original 24 unsupported-call assertions failed before the change. The final focused tests pass all 104 assertions; all upstream-registered JSON Schema tests pass (5,726 assertions, 40 cases) under GCC 15 in C++17 with `-Wall -Wextra -Werror`. The new test file also passes a C++11 syntax-only check, and direct unsupported class instantiations produce the expected diagnostics.

AI assistance: prepared with OpenAI Codex, with reproduction and verification performed locally.

### DIFF
--- a/doc/ref/jsonschema/json_schema.md
+++ b/doc/ref/jsonschema/json_schema.md
@@ -12,6 +12,14 @@ A `json_schema` is immutable and thread-safe.
 
 The class satisfies the requirements of MoveConstructible and MoveAssignable, but not CopyConstructible or CopyAssignable.
 
+`Json` must be a `basic_json` specialization with `char_type` equal to `char`
+and a default-constructible allocator. For `std::scoped_allocator_adaptor`,
+the outer allocator and all inner allocators must be default constructible.
+This includes `json`, `ojson`, and their `jsoncons::pmr` variants. Wide-character
+types such as `wjson` and `wojson`, and allocators without a default constructor,
+are not supported. Instantiating `json_schema` with these unsupported types
+produces a compile-time diagnostic.
+
 #### Member functions
 
 <table border="0">
--- a/doc/ref/jsonschema/jsonschema.md
+++ b/doc/ref/jsonschema/jsonschema.md
@@ -3,6 +3,12 @@
 Since 0.174.0, the jsonschema extension implements Drafts 4, 6, 7, 2019-9 and 2020-12 of the [JSON Schema Specification](https://json-schema.org/specification).
 Previous versions supported Draft 7.
 
+The extension requires a `basic_json` type with `char_type` equal to `char`
+and a default-constructible allocator, including any underlying scoped allocators.
+It supports `json`, `ojson`, and their `jsoncons::pmr` variants, but not
+wide-character types such as `wjson` and `wojson` or allocators without a default
+constructor. See [json_schema](json_schema.md) for the type requirements.
+
 The documentation below describes the new features for the jsonschema extension since 0.174.0.
 For earlier releases, please refer to [jsonschema (until 0.174.0)](https://github.com/danielaparker/jsoncons/tree/main).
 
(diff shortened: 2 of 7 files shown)


danielaparker:
Thanks for the contribution! especially for the test coverage. The two failing checks are unrelated to your code, so not your concern. 

I do think there's some verbosity in the code to impose the restrictions on `make_json_schema`, there's more code than I think is needed. We already have some traits structs for imposing restrictions on allocators and char types, the use of which would help with uniformity and maintainability throughout the library. When I have time I'll review this more carefully.

Alb3e3:
I checked the existing traits against the constraints here. Two details may be useful for your review:

- `ext_traits::is_narrow_character` also accepts `char8_t`. I tried substituting it locally, and `make_json_schema(basic_json<char8_t>(true))` then fails inside the schema factory on its narrow string operations.
- `ext_traits::is_propagating_allocator` accepts both supported scoped allocators and the non-default-constructible allocator from `mock_stateful_allocator.hpp`. On libstdc++, `std::is_default_constructible` also reports true for that scoped wrapper, which is why the helper checks its underlying allocators.

I'll wait for your closer review before changing the constraints. The tests should make it straightforward to check a simpler formulation while preserving those cases.

danielaparker:
Apologies, just getting back to this. I believe the constraints that the Json type must currently satisfy for `make_json_schema` are just

`std::is_same<typename Json::char_type,char>::value` must be true

Either
`jsoncons::ext_traits::is_polymorphic_allocator<typename Json::allocator_type>::value` 
or 
`std::allocator_traits<std::allocator<typename Json::allocator_type>>::is_always_equal::value`
must be true

That should be it.

Issue #740, opened 2026-09-14
JMESPath - operator in lexical scope


BartelNieuwenhuyse:
I found what I think is a precedence bug in jmes path expression

The following expression:

```let $f = foo, $b = bar in $f || $b```

gives an _unknown variable_ error on evaluation of $b. I think because the or operator consumes the complete let .. in expression as its left hand side, such that the variables $f and $b have gone out of scope when we enter the right hand side of the or expression.

But even parenthesizing does not help:

```let $f = foo, $b = bar in ($f || $b)```

gives the same error.

It is not clear to me from the specs what the correct precedence should be: should the unparenthesized expression yield the same as the parenthesized expression above (I guess it should), or should it act as the following expression: 

```(let $f = foo, $b = bar in $f) || $b```

in which case $b is rightly not known on the right of the or operator. 

**What jsoncons library version?**

 - [ ] Latest release 1.9.0
 - [ ] Other release ______
 - [x] master

PR #744, opened 2026-09-18
Bugfix/jmespath operator in let evaluation expression

BartelNieuwenhuyse:
Issue #740 : 

The following expression:

```let $f = foo, $b = bar in $f || $b```

gives an unknown variable error on evaluation of $b.


danielaparker:
Thanks!

PR #745, opened 2026-09-19
Include missing <exception> for !defined(JSONCONS_NO_EXCEPTIONS)

kou:
We use `std::terminate()` with
`!defined(JSONCONS_NO_EXCEPTIONS)`. `std::terminate()` requires `#include <exception>`.


danielaparker:
Thanks for contributing!

